PR #7, opened 2019-07-23
Move LDAS to FV_PRECISION=R4

mathomp4:
I don't think the LDAS needs to use the R4R8 workaround. Set to be R4


mathomp4:
Note. This is a test of the `CODEOWNERS` file. As @pchakraborty is not in @GEOS-ESM/cmake-team, he is the "non-CMake" person I wanted to test this with. It does look like the @GEOS-ESM/cmake-team was triggered in the required reviewers.

PR #37, opened 2019-09-05
add optional average sun angle step size

weiyuan-jiang:


weiyuan-jiang:
This change will lead to non-zero diff

PR #45, opened 2019-09-13
Add FMS as a new external

mathomp4:
To use GMAO_Shared 1.0.6 and higher, you need to add FMS as a new
external. This means *every* `manage_externals` `.cfg` file needs to
move up because of the change to `src/Shared/CMakeLists.txt`

### DIFF
--- a/Develop.cfg
+++ b/Develop.cfg
@@ -28,6 +28,13 @@ local_path = ./src/Shared/@MAPL
 branch = master
 protocol = git
 
+[FMS]
+required = True
+repo_url = git@github.com:GEOS-ESM/FMS.git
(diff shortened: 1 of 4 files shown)


mathomp4:
@JustinPerket, note that hopefully with MAPL 2.0 we can un-require FMS because the FV3 dependency will be gone.

JustinPerket:
Tested to build, looks good to merge. 0-diff testing for ```jperket/PrepForMasterMerge``` will occur before merging to develop, but don't anticipate any issues

https://github.com/GEOS-ESM/GMAO_Shared/issues/40 can be marked as closed

PR #47, opened 2019-09-18
Release/v17.8.0

JustinPerket:
Change to externals. 

Now with merge https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/77 , tag [ldas.2](https://github.com/GEOS-ESM/GEOSgcm_GridComp/releases/tag/ldas.2) was created.

[feature/ldas-surface](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/ldas-surface) will no longer be the default in Develop.cfg (and tested with the nightly test). Features from this branch not in develop are still planned to be incorporated, but due to the work required, we can't wait. [feature/ldas-surface](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/feature/ldas-surface) will be considered semi-frozen soon, and new GEOSgcm_GC branches for LDAS development should made from [develop](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop)

Though we're not ready for a final release version, will also do a PR to master following [these guidlines for release branches](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow). 

Update: that PR is #48

Issue #21, opened 2019-07-31
documentation update needed


JustinPerket:
1. Update readme's and tutorial for git

    - Ex, different directory structure. placement of exe and bat files, etc    

2. Update readme's and tutorial for m4_17-8 updates

JustinPerket:
> Update readme's and tutorial for git
#34 address this

JustinPerket:
From: https://github.com/GEOS-ESM/GEOSldas/pull/48#issuecomment-532769547 and @gmao-rreichle, we could stand to convert the pptx tutorial to something that will work with version control.
Perhaps we should develop the LDAS github wiki? There are other readme files (like history and metforcing) that could be converted to markdown / wiki pages instead.

 
> > Ew. Binary PPTX. @tclune we should probably talk about things like this entering Git. It's already there so...
> 
> Matt: I agree that this is suboptimal from a repository standpoint. But we do need some sort of presentation-quality material for tutoring the land group on GEOSldas. Having this material outside of git makes it even more difficult to keep it in sync with GEOSldas. Let me know if Tom has an issue with this, or perhaps a better solution.

JustinPerket:
look through the tutorial, there's nothing that couldn't be converted to markdown if we wanted. Tables would be fine. If there were images, that wouldn't be a problem either. These version control pages could be the go-to place for information. The actual slides that we present don't need to be under version control, if it's duplicating/condensing information that would already be accessible through a wiki. In fact, it would be redundant, and increasing the risk we update information one document, but forget to update the other(like the powerpoint and README's now)

JustinPerket:
We now have a very simply wiki, that links to docs in master:

https://github.com/GEOS-ESM/GEOSldas/wiki/GEOSldas-wiki

gmao-rreichle:
Thanks for putting together the GEOSldas Wiki, Justin.  Very helpful.
A couple of minor things after a quick glance:
1) For non-experts (incl myself), the "clone" line should probably have a place-holder entry for the tag (or branch?).  
2) We probably shouldn't say "GMAO Desktops" -- I doubt this works on my Windows Desktop.  I think you mean "GMAO Linux systems" (which includes Linux desktops and shared servers).  Not sure if it works on Macs.

gmao-rreichle:
[...]

PR #48, opened 2019-09-18
Release/v17.8.0

JustinPerket:
This PR is occuring with https://github.com/GEOS-ESM/GEOSldas/pull/47

> Though we're not ready for a final release version, will also do a PR to master following [these guidlines for release branches](https://www.atlassian.com/git/tutorials/comparing-workflows/gitflow-workflow)


gmao-rreichle:
> Ew. Binary PPTX. @tclune we should probably talk about things like this entering Git. It's already there so...

Matt: I agree that this is suboptimal from a repository standpoint.  But we do need some sort of presentation-quality material for tutoring the land group on GEOSldas.  Having this material outside of git makes it even more difficult to keep it in sync with GEOSldas.  Let me know if Tom has an issue with this, or perhaps a better solution.

JustinPerket:
> > Ew. Binary PPTX. @tclune we should probably talk about things like this entering Git. It's already there so...
> 
> Matt: I agree that this is suboptimal from a repository standpoint. But we do need some sort of presentation-quality material for tutoring the land group on GEOSldas. Having this material outside of git makes it even more difficult to keep it in sync with GEOSldas. Let me know if Tom has an issue with this, or perhaps a better solution.

This seems relevant to https://github.com/GEOS-ESM/GEOSldas/issues/21 , so we can use that to centralize documentation / tutorial updates we want. 

The mpi command now also has an issue, since need to consult with @weiyuan-jiang : https://github.com/GEOS-ESM/GEOSldas/issues/49

For now, I will merge so we can have a working master

JustinPerket:
look through the tutorial, there's nothing that couldn't be converted to markdown if we wanted. Tables would be fine. If there were images, that wouldn't be a problem either. These version control pages could be the go-to place for information. The actual slides that we present don't need to be under version control, if it's duplicating/condensing information that would already be accessible through a wiki. In fact, it would be redundant, and increasing the risk we update information one document, but forget to update the other(like the powerpoint and README's now)

PR #50, opened 2019-09-18
Justin perket/update readme

JustinPerket:
remove erroneous reference to AGCM


JustinPerket:
fair enough, will do in the future. it was just obviously wrong

Issue #51, opened 2019-09-18
Reminder to put par bug fix and zenith angle time stepping fix into develop


JustinPerket:
but not for master/ v17.8.0 release

PR #52, opened 2019-09-19
Feature/sun angle stepsize

JustinPerket:
resuming this pull request of @weiyuan-jiang's. First making it a draft pull request to alert @gmao-rreichle, and allow @weiyuan-jiang to review these are the correct changes.

Will do system tests, and other tests as needed for acceptance 

This is 1/2 of #51

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/GEOS_MetforceGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/GEOS_MetforceGridComp.F90
@@ -948,6 +948,7 @@ contains
             currTime=internal%mf%TimePrv,                                       &
             INTV=internal%mf%ntrvl,                                             &
             ZTHB=internal%mf%zenav,                                             &
+            STEPSIZE=150.0,                                                     &
             rc=status                                                           &
             )
        VERIFY_(STATUS) 
@@ -980,6 +981,7 @@ contains
          INTV=ModelTimeStep,                                                    &
          ZTHB=zth,                                                              &
          currTime=ModelTimeCur,                                                 &
+         STEPSIZE=150.0,                                                        &
          rc=status                                                              &
          )
     VERIFY_(status)


JustinPerket:
I have runs for global and cubed-sphere global with testing this. It's 5-min frequency. Initailly I ran for an hour, but then extended the run to a day so you can see the solar differences.

A quick comparison of lfs show SWdown differences up to ~0.5 W/m2. The spatial differences shift from being predominately on one side of the  terminator at the beginning of an hour, to being predominately on the other at the end of an hour. This pattern repeats hourly. 

Here is a quick plot showing this, SWdown difference, Control - Test :
image


I should add that though hard to see on the plot, there are differences of ~0.2 W/m2 across sunlit areas, away from the terminator.

gmao-rreichle:
I looked at Justin's 24-h test runs with 5-min timestepping & output ("GlobalDevelop" and "GlobalTest.0").   There are large differences in a couple of tiles in New Zealand (DeltaTSURF>6 K).  Inspection of time series suggests that these differences are due to numerical instabilities in the model.  The output for these tiles doesn't make sense even before the two runs diverge, which is probably because of the bugs in the "Icarus-NL" bcs.  When you pick up this work again, please re-run the tests with "Icarus-NLv3" bcs.

gmao-rreichle:
PS: Please do not overwrite the existing output.  I will want to compare the results across the different bcs so we can also update the bcs in the nightly test in the same non-zero-diff change.

gmao-rreichle:
For reference, here's an example of a crazy time series for a tile in New Zealand from the earlier experiments with the original "NL" bcs:
/home/jperket/project/ldas/GlobalDevelop
/home/jperket/project/ldas/GlobalTest.0
tmpplotseries_cell3_SFMC

Here's the corresponding result from the most recent runs with "NLv3" bcs:
Develop, control:             /home/jperket/project/ldas/DevGlobal_INLv3/
Zenith step fix +PAR fix:  /home/jperket/project/ldas/ZenParGlobal_INLv3/
tmpplotseries_cell3_SFMC

This tells me that the original NL bcs occasionally result in numerically unstable simulations and therefore cannot be used for testing of non-zero diff runs.

An analysis of global maps and difference maps (not shown) for the final 5-min time step of the above runs confirm that the zenith angle change is ok.

However, before I approve this pull request, I would still like to see results from the SMAP assimilation test case with the zenith angle change and the PAR perturbations fix.

PR #56, opened 2019-10-10
@gmao-rreichle updated LDAS_Forcing.F90

JustinPerket:
This updated LDAS forcing to use 522_fp data. I labeled hotfix because it's needed quickly for 17.8.0 (though technically not a bug).  After this Pull request to develop, it should soon get into master, and a new release tag should be made.


JustinPerket:
pretty clearly 0-diff, tested to work (with FPP spinup runs). Merging

PR #61, opened 2019-11-20
speed up reading bin. fix writig nml (root only)

weiyuan-jiang:
it speeds up initialization ( reading bin in continuous array, writing nml only in root process)


weiyuan-jiang:
Sure. Will change to -9999 to make it consistent across the codes

PR #68, opened 2019-12-04
Tomerge/develop smap

JustinPerket:
This has all 0-diff changes of @weiyuan-jiang's branch `develop_smap`, but temporarily remove non-0diff zenith stepsize arguments in GEOS_MetforceGridComp.F90. After nightly test passes, reinsert the stepsize to solidly prove those are the only non-0diff changes.

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -9,7 +9,7 @@ protocol = git
 required = True
 repo_url = git@github.com:GEOS-ESM/ESMA_cmake.git
 local_path = ./@cmake
-tag = v1.0.9
+tag = v1.0.11
 externals = Externals.cfg
 protocol = git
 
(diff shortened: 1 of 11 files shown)


gmao-rreichle:
@JustinPerket 
I think the zenith step size changes aren't the only non-zero diff changes.
Isn't the call to check_catch_progn() another non-zero-diff change?
Maybe you also removed this already and only the intro comments for this PR were inaccurate.
If the call to check_catch_progn() remains in the merge, I would expect the nightly test to fail.

JustinPerket:
> @JustinPerket
> I think the zenith step size changes aren't the only non-zero diff changes.
> Isn't the call to check_catch_progn() another non-zero-diff change?
> Maybe you also removed this already and only the intro comments for this PR were inaccurate.
> If the call to check_catch_progn() remains in the merge, I would expect the nightly test to fail.

You're right. Will fix now before the nightly test.

PR #91, opened 2019-12-18
Dev/mapl 2.0

weiyuan-jiang:
1) use MAPL 2.0
2) rm fvcore
3) add cs forcing and data assimilation


weiyuan-jiang:
The branch checkout gcm_GridComp dev/MAPL-2.0 which is different from develop. Ben told me he merged develop to dev/MAPL2.0 frequently.

mathomp4:
@weiyuan-jiang Does this mean that GEOSldas does okay with Intel MPI? We are still using that as default in MAPL 2.0's `@env`.

weiyuan-jiang:
I plan to switch to impi which will make less trouble for the baselib. The initial tests seem ok. I will see how it is going tonight

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -884,6 +867,10 @@ subroutine Initialize(gc, import, export, clock, rc)
        out_smapL4SMaup,                         &
        N_obsbias_max                            &
        )
+      call MAPL_GetResource ( MAPL, GridName, Label="GEOSldas.GRIDNAME:", DEFAULT="EASE", RC=STATUS)
+      VERIFY_(STATUS)
+      if (index(GridName,"-CF") /=0) out_smapL4SMaup = .false. ! no out_smap for now if it is cs frid
```
gmao-rreichle: maybe fix typo in comment: "grid" in instead of "frid"

JustinPerket:
> @weiyuan-jiang Does this mean that GEOSldas does okay with Intel MPI? We are still using that as default in MAPL 2.0's `@env`.

It would be great to here that we no longer take a speed hit from this.

src/Components/GEOSldas_GridComp/Shared/LDAS_TileCoordType.F90
```diff
@@ -76,10 +76,15 @@ module LDAS_TileCoordType
      real    :: max_lat    ! maximum latitude (bounding box for tile)
      integer :: i_indg     ! i index (w.r.t. *global* grid that cuts tiles) 
      integer :: j_indg     ! j index (w.r.t. *global* grid that cuts tiles)
+     !if it is Cubed-Sphere grid, the index will be saved here for forcing
```
gmao-rreichle: I'm a bit confused by the naming of the field here. 
In my understanding the fields %i_indg and %j_indg are w.r.t. to the global grid that cuts tiles (regardless of whether that is a cube-sphere, EASE, or lat-long grid)
The second set of fields %cs_i_indg and %cs_j_indg should be w.r.t. to the (lat/long) "hash" grid that is used to quickly identify tiles that are within the lat/long grid cell(s) of the "hash" grid.  The name "cs_[x]_indg" is confusing, because the grid in question is NOT a cs grid.  How about something like:  %hash_i_indg and %hash_j_indg ???

src/Components/GEOSldas_GridComp/Shared/LDAS_ensdrv_mpi.F90
```diff
@@ -208,6 +208,8 @@ subroutine init_MPI_types()
     !     real    :: max_lat    ! maximum latitude (bounding box for tile)
     !     integer :: i_indg     ! i index (w.r.t. *global* grid that cuts tiles) 
     !     integer :: j_indg     ! j index (w.r.t. *global* grid that cuts tiles)
+    !     integer :: cs_i_indg  ! i index (w.r.t. *global* grid that cuts tiles) 
+    !     integer :: cs_j_indg  ! j index (w.r.t. *global* grid that cuts tiles)
```
gmao-rreichle: See comment re. LDAS_TileCoordType.F90

gmao-rreichle:
Approving pull request today, but we still need to 
1) add proper documentation in change log file, and 
2) at a later time, address the conversion of the "%cs_i_indg" and "%cs_j_indg" into separate variables (arrays) in assim GridComp.

PR #98, opened 2020-01-16
no check for cat_progns after perturbations

weiyuan-jiang:
remove the check for cat progns after perturbations


gmao-rreichle:
Thanks for putting together the pull request.
It would perhaps be better to comment out the lines in question, rather than remove them entirely.  In this way, it will later be easier to follow or re-instate, should that be needed.  But perhaps that's just my inexperience with git speaking.

In addition to the changed lines, don't we also want to comment out (or delete) Lines 45-47?:

  !use catch_types, only: cat_param_type
  !use catch_incr, only: check_catch_progn
  !use GEOS_EnsGridCompMod,      only: cat_param=>catch_param

weiyuan-jiang:
Good catch! I have commented those lines out too. Some (including me) prefer to delete those codes to make the program clean.  Since the codes can alway be retrieved, we won't lose them. But I agree it is more convenient to bring them back if they are simply commented out.

Issue #122, opened 2020-02-21
add 'catch_progn_incr'  to HISTORY collection 


saraqzhang:
'catch_progn_incr'  holding ldas_incr variables should be added to GEOSldas_HIST.rc.
 It should be un-commented when in the coupling configuration ( an separate issue on  GEOSldas_LDAS.rc and lenkf.j.template ) .

gmao-rreichle:
Sara, 
Please go ahead and create a pull request based on the latest "develop" branch.
I plan to further edit GEOSldas_HIST.rc (add a minor comment) once the pull request gets to me for approval.  
- Rolf

PR #124, opened 2020-02-24
     add catch_progn_incr to HISTORY collection

saraqzhang:
issue #122


saraqzhang:
the incr output is .bin from History, only be converted by tile_bin2nc in post process.

saraqzhang:
the incr file name (.bin) by History.rc  is seen by post processing tile_bin2nc,  it is not hard-wired in GCM interface.

saraqzhang:
the incr output is .bin from History, only be converted by tile_bin2nc in post process.
the incr file name (.bin) by History.rc is seen by post processing tile_bin2nc, it is not hard-wired in GCM interface.

gmao-rreichle:
I deleted the redundant copy of Sara's comments.

Re. the ".bin" template for "catch_progn_incr" -- sorry, I keep forgetting that we can't write tile-space output directly to nc4.  Let's leave ".bin" as is.

Should/could we still change the Collection name for "catch_progn_incr"?

saraqzhang:
@gmao-rreichle  the name can be changed if to make it more inline with history convention.

saraqzhang:
@weiyuan-jiang @gmao-rreichle 
it seems that process_hist.csh needs to be modified accommodating the addition/modification to History.rc here.

weiyuan-jiang:
We can change the resolution to 576-by-361 simply by replacing the 720-by-361 label

gmao-rreichle:
- fixed "tavg1_*" in collection names (my mistake in earlier round of editing)
- reordered collection names to match order in which collections are defined below

gmao-rreichle:
@saraqzhang , regarding your comment:

> it seems that process_hist.csh needs to be modified accommodating the addition/modification to History.rc here.

I suppose you're referring to the addition of the "catch_progn_incr" collection?  If so, please add the necessary changes to process_hist.csh on this branch.  But I'm not entirely sure I really understand what else needs to be done.  There isn't (yet) an automated test case that would require output of the "catch_progn_incr" collection.  I had thought that adding "catch_progn_incr" to GEOSldas_HIST.rc was simply a good place to put the collection definition for subsequent manual setup/config of experiments that need this output (through editing of the HISTORY.rc file that's generated by process_hist.csh).  I'm probably missing something.

gmao-rreichle:
Merging these changes now to lock in some progress.
At a later time, we need to make further changes that will allow users to see a more complete HISTORY.rc template regardless of the choices in process_hist.csh.

Issue #123, opened 2020-02-24
GEOSldas "log" file


gmao-rreichle:
Make sure a "log" file is available from each GEOSldas run.
From Weiyuan's email of 24 Feb 2020:
"slurm has problem to produce the log file in time.  In our GEOEldas, we move and rename the log file to the location we specified. But sometimes, when we move, the file is not there yet, or, it is still being produced"

saraqzhang:
@weiyuan-jiang @gmao-rreichle :   it  is a  bug in lenkf.j.template ( in the new develop branch) :  line 78 :  /bin/rm  -rf  *  .  It removes everything in scratch/ including the log.txt and err.txt that are just generated and put there by slurm. so during the run we end up without log and err from the operation. 
I made test run with line 78 changed to  /bin/rm -rf  *.*  , that guaranteed log and err files not removed and in place to record run messages as expected, and later on be moved to rc_out.

weiyuan-jiang:
OK. That reminds me of something. It was rm -f  *.*  before . That is why I chose the log file name without dot "." to avoid being deleted.   Somehow, I drop this by accident. But we have better solution now. The log file will be  generated in the right place.

gmao-rreichle:
addressed in pull request #127

Issue #112, opened 2020-02-14
Compression of pert restart files


gmao-rreichle:
Pert restart nc4 files used to be compressed internally but that proved too slow, so the compression was removed with the merger of dev/MAPL-2.0

Pert restart files without compression are too large for production runs.  Need to figure out how to get some sort of compression back, perhaps simple gzip after the fact.   TBD

gmao-rreichle:
Use zip to compress all restarts EXCEPT those at the final time (which will typically be needed again in short order)
Assume in the setup script that restarts are not zipped.
If users want to restart from a zipped set of restarts, they need to unzip (after making a copy if needed).
It is not helpful to have the script unzip restarts because that just creates file writing permission issues (if someone is using zipped restarts from a run by another user)

gmao-rreichle:
addressed in pull request #127

Issue #110, opened 2020-02-14
SLES12


gmao-rreichle:
Need GEOSldas branch and test using SLES12 on Discover.

Need to understand if SLES12 requires two tags, one for SLES11 and another for SLES12.

gmao-rreichle:
addressed in pull request #127

Issue #109, opened 2020-02-14
Update Fortran and CMakeLists.txt to be MAPL 2.0 consistent


mathomp4:
One thing that should be done is when MAPL 2.0 is used, in Fortran files:
```fortran
use MAPL_Mod
```
should become:
```fortran
use MAPL
```

And in `CMakeLists.txt` files, dependencies or references to `MAPL_Base` should become `MAPL`.

This should be pretty simple search-and-replace, but it's possible this can expose some other underlying issue (some hidden dependency).

gmao-rreichle:
Addressed in pull request #127

PR #127, opened 2020-02-25
add SLES 12 support and compress land pert restart

weiyuan-jiang:
#123 #112 #110 #109 issues are addressed in this pull request


gmao-rreichle:
@mathomp4 : This pull request would use v2.0.1 of ESMA_env, which points to the Intel 18 compiler, which works for us but produces an error message.
@weiyuan-jiang is telling me that you are about to issue a new release of ESMA_env that would move to Intel 19 or 20 and thereby fix the issue with the error message.
Should we wait for you to do that, and then update the pull request accordingly?  
We do need to create a new GEOSldas tag this week at the very latest.

mathomp4:
@gmao-rreichle I'm currently testing new `g5_modules` setups. I had thought about moving to Intel Compiler 20 with Intel MPI 20, but that is a non-zero-diff change. But I asked NCCS to allow Intel Compiler 18 with the later Intel MPI versions. They fixed up lmod and I'm currently building an Intel 18.0.5 + Intel MPI 19.1 Baselibs. Hopefully this will allow SLES11 and SLES12 versions of GEOSgcm (and ldas?) to be zero-diff between the two machines.

gmao-rreichle:
@mathomp4 : I think @weiyuan-jiang tested Intel Compiler 20 (presumably with Intel MPI 20) and found it to be zero-diff for GEOSldas.  So I assume that whatever you get to work for the GCM should also work for GEOSldas.  Please keep us posted.

mathomp4:
@gmao-rreichle My guess is that the non-zero-diff ness of the GEOSgcm comes from the vectorization we turn on in FV3. The rest of the model is compiled pretty safely (precise math) so it's rare to have a compiler-induced difference. But not impossible...

I'm currently building GEOSgcm and will test soon. If it works, I'll pass on a `g5_modules` for @weiyuan-jiang to try.

mathomp4:
Okay. I think a combination of Intel Compiler 18 and Intel MPI 19.1 seems to work in my testing. Before I commit it to ESMA_env, I've asked @weiyuan-jiang (on Skype, but we can track it here too) to test:
```
/gpfsm/dswdev/mathomp4/Models/GEOSgcm-stock-SLES11_and_12/GEOSgcm/@env/g5_modules
```
with the LDAS. If that works, I'll make a new release.

mathomp4:
@weiyuan-jiang said that my new modules setup works. I have made a new release of ESMA_env [v2.0.2](https://github.com/GEOS-ESM/ESMA_env/releases/tag/v2.0.2)

Issue #147, opened 2020-03-11
bug fix: correct log file names in case SEGMENTS are changed after ldas_setup


gmao-rreichle:
The fix for missing log file output that went into the "beta.2" tag does not work when SEGMENTS are changed after ldas_setup.  Restore this flexibility by dealing with log file names in post-processing.

gmao-rreichle:
Addressed with #150

PR #150, opened 2020-03-17
reinstate move log files

weiyuan-jiang:
1) move log file after each job
2)link to original cat_param and mwrtm param files


weiyuan-jiang:
address #147

PR #162, opened 2020-03-27
Update README.md

weiyuan-jiang:


gmao-rreichle:
@weiyuan-jiang : It's a bit unclear to me what the revised sentence means exactly:

> If you are using SLES12 at NCCS, you **must** run setup on an interactive compute node 
> and source g5_modules.sh under bash shell. 

Does this mean that for SLES12 at NCCS
a) "source g5_modules" is *only* needed under bash (and not needed if tcsh is used)?
or 
b) setup can *only* be run with bash (and does not work with tcsh)?
If I know whether it's a) or b), I can try to rephrase.
- Rolf

weiyuan-jiang:
Under csh, we need to source g5_modules.  Under interactive node, usually the default is bash shell ( at least mine is bash), so we need to source g5_modules.sh

gmao-rreichle:
Sorry, still not clear...
> Under csh, we need to source g5_modules. Under interactive node, usually the default is bash shell 
> (at least mine is bash), so we need to source g5_modules.sh

First, I'm reading that the commands are different under csh and bash:
`source g5_modules` (csh)
`source g5_modules**.sh**` (bash)

This seems separate from the second issue, which is that *you* are using bash, but others might be using csh.  

Maybe this is best left for discussion at our next tag-up meeting.

weiyuan-jiang:
Maybe I should not mention interactive mode that causes confusion. What matter is the shell
source g5_modules (csh)
source g5_modules.sh (bash)

PR #180, opened 2020-04-16
Use ESMA_env with Intel-19 compiler

gmao-rreichle:
Use ESMA_env with Intel-19 compiler (v2.0.4+intel19.10).
@weiyuan-jiang tested before with Intel-19 and found zero-diff.
Assuming that this will be ok.  Automated tests should confirm tonight.


weiyuan-jiang:
1) This will only work for sles12. There is only one choice in g5_modules.
2) The night test is still on sles11. So it will fail.

PR #182, opened 2020-04-21
checkpoint files are geneareted by GEOSldas 

weiyuan-jiang:
This pull request addresses two things:
- Don't need the line directing the landpert checkpoint output when on cube-sphere tiles
- Create directories from within F90 executable for ObsFcstAna files.  (Facilitates extending an existing GEOSldas run without going through setup.)  This will be revamped later on so that all files are created in ./scratch and then cleaned up in post-processing.

### DIFF
--- a/src/Applications/LDAS_App/ldas_setup
+++ b/src/Applications/LDAS_App/ldas_setup
@@ -1180,7 +1180,7 @@ class LDASsetup:
                 valn = self.catch+ensid+'_internal_checkpoint'
                 ldasrcInp[keyn]= valn
 
-                if((self.has_ldassa_pert or self.has_geos_pert) and _perturb == 1):
+                if('-CF' not in self.rqdExeInp['GRIDNAME'] and _perturb == 1):
                     keyn = 'LANDPERT_INTERNAL_CHECKPOINT_FILE'
                     valn = 'landpert'+ensid+'_internal_checkpoint'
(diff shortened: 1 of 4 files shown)


weiyuan-jiang:
We can replace the non-standard "system" call with a standard "excute_command_line" call.  That should not cause problems.

gmao-rreichle:
Not sure how much protection is provided by "execute_command_line()"
It looks like it's only a wrapper for the "system()" call, and it may not be available in all compilers.
https://gcc.gnu.org/onlinedocs/gfortran/EXECUTE_005fCOMMAND_005fLINE.html
https://www.pgroup.com/userforum/viewtopic.php?t=6422
We could use it as a temporary solution.
In the long run, shouldn't we handle the files in `ana/` just like HISTORY files, i.e., write them into `scratch/` and clean up in post-processing?
(Note that the subroutines in "clsm_ensupd_read_obs.F90" that still contained "system()" calls are not currently in use but need to be overhauled before the land analysis can be integrated into the next GMAO reanalysis.)

weiyuan-jiang:
This intrinsic call is Fortran 2008 standard. So for the long run, it should be fine. But I don't oppose to send data from scratch to the destination. I just need more information to understand the file name format and where to send. The original codes have dealt with all the cases that I may never see in the limited test cases. So it is much easier to change the original codes.

Issue #136, opened 2020-02-26
add forecast error covariance inflation


gmao-rreichle:

gmao-rreichle:
Addressed with pull request #203

PR #203, opened 2020-05-01
Add forecast error covariance inflation

gmao-rreichle:
Addresses #136


gmao-rreichle:
Verified and tested by @gmao-qliu (email of 1 May 2020):

>With the default factor value (-9999). The branch passed nightly test with -t global on both sles11 and sles12. 
>With factor value of 1.2, I get the exactly same result as my version 
"feature/QL_cov_inflation_upd".  
>I also ran the ctrl and cov_1 tests, all results are good and consistent.

PR #207, opened 2020-05-04
fix mpi_finalize fail

weiyuan-jiang:
Change ESMA_env and MAPL version

### DIFF
--- a/Externals.cfg
+++ b/Externals.cfg
@@ -2,7 +2,7 @@
 required = True
 repo_url = git@github.com:GEOS-ESM/ESMA_env.git
 local_path = ./@env
-tag = v2.1.2+intel19.1.0
+tag = v2.1.3+intel19.1.0
 protocol = git
 
 [ESMA_cmake]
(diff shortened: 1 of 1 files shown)


mathomp4:
Hold please. Matt was an idiot

mathomp4:
Yeah. Matt tagged `v2.1.3` with `v2.1.2+intel19.1`. I created a `v2.1.3+intel19.1` and pushed that tag. It is *identical* to `v2.1.2+intel19.1` I was just being clumsy/stupid.

mathomp4:
Ah. I figured it out. The LDAS never got `v2.1.2` since it was a slight NAS bugfix. LDAS doesn't run at NAS as far as I know, so I never gave you one. So, when I was making this new ESMA_env tag, I saw the last I gave you was 2.1.1 and said, "Let's advance that patch by 1". 

🤞 this is the last time I do this. Sorry about that

gmao-rreichle:
@weiyuan-jiang : Can you please update the branch if needed and then re-test?  I'm a bit lost as to what exactly the tags should be in Externals.cfg.

weiyuan-jiang:
I have updated it and am testing it fully under sles12

PR #212, opened 2020-05-06
Feature/diagAandB

saraqzhang:
5x2( fcst & ana)  diag variables are collected into a HISTORY.rc collection :catch_diag_aup. 
the system is based on develop branch as 4/30/2017.


src/Applications/LDAS_App/GEOSldas_HIST.rc
```diff
@@ -17,6 +17,7 @@ COLLECTIONS:
 #CUBE            'tavg24_2d_lnd_Nx'
 #ASSIM           'SMAP_L4_SM_gph'
 #                'catch_progn_incr'
+#                'catch_diag_aup'
```
gmao-rreichle, line 20: We still need to clean up HISTORY conventions...
One concern is that it's probably impossible to keep the Collection label to three letters, but we can still indicated whether this is tavg or incr, what the time step is, and whether the output is in tile space.
For this new collection I suggest using:   "inst3_1d_[LABEL]_Nt" for instantaneous, 3-hourly, 1-dim increments output in tile space.
For label, perhaps use "lndfcstana"?  It's a bit misleading because we're not putting out everything that is in the "lnd" collection, although anything that would be in a inst3 lnd collection other than the 5 variables included here would represent both fcst and ana values (which are identical for our typical analysis use case).

src/Applications/LDAS_App/GEOSldas_HIST.rc
```diff
@@ -359,3 +360,27 @@ COLLECTIONS:
...
+  catch_diag_aup.regrid_exch: '../input/tile.data',
+  catch_diag_aup.regrid_name: 'GRIDNAME', 
+  catch_diag_aup.grid_label:  PC720x361-DC,
+  catch_diag_aup.deflate:     2,
+  catch_diag_aup.fields:     'SFSM_ANA'    , 'LANDASSIM'      ,
+                              'RZSM_ANA'    , 'LANDASSIM'      ,
+                              'PFSM_ANA'    , 'LANDASSIM'      ,
+                              'TPSURF_ANA'    , 'LANDASSIM'      ,
+                              'TSOIL1_ANA'    , 'LANDASSIM'      ,
+                              'SFSM_FCST'    , 'LANDASSIM'      ,
```
gmao-rreichle, line 379: I think we can delete the five *_fcst variables throughout. 
They should be identical to using: 
      'WCSF' , 'ENSAVG'  , 'SFMC_ANA'
and similarly for the other variables.  The third column specifies the name of the variable in the nc4 file.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -716,9 +719,110 @@ subroutine SetServices ( GC, RC )
...
+       ! 1: sm_surface_[forecast/analysis]           [m3 m-3]
+       ! 2: sm_rootzone_[forecast/analysis]          [m3 m-3]
+       ! 3: sm_profile_[forecast/analysis]           [m3 m-3]
+       ! 4: surface_temp_[forecast/analysis]         [K]
+       ! 5: soil_temp_layer1_[forecast/analysis]     [K]
+
+    call MAPL_AddExportSpec(GC,                    &
+    LONG_NAME          = 'surface_soil_moisture_analysis'      ,&
+    UNITS              = 'm3 m-3'                         ,&
+    SHORT_NAME         = 'SFSM_ANA'                      ,&
```
gmao-rreichle, line 732: The SHORTNAME here should be either WCSF_ANA or SFMC_ANA.
WCSF is consistent with the variable and export names in the F90 file through the LANDASSIM, LAND, and CATCH GridComps.
SFMC would be consistent with the variable name in the published lnd Collection.
Perhaps it's more consistent (and cleaner in terms of the F90) to use WCSF_ANA and then put the following line into the HISTORY.rc file:
    'WCSF_ANA'    ,  'LANDASSIM'  ,  'SFMC_ANA'
(Similarly for the other variables.)
But then again, the problem with WCSF is that the _long_ names are ridiculous from the perspective of a land modeler and urgently need cleaning up, but that's a known issue with published GEOS file specs.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -716,9 +719,110 @@ subroutine SetServices ( GC, RC )
...
+    LONG_NAME          = 'soil_temperatures_layer_1_analysis' ,&
+    UNITS              = 'K'                         ,&
+    SHORT_NAME         = 'TSOIL1_ANA'                ,&
+    DIMS               = MAPL_DimsTileOnly           ,&
+    VLOCATION          = MAPL_VLocationNone          ,&
+                                           RC=STATUS  )
+     VERIFY_(STATUS) 
+
+!--
+    call MAPL_AddExportSpec(GC,                    &
```
gmao-rreichle, line 775: i don't think the *_fcst export variables are needed, see comment above

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1718,10 +1876,18 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
 
        if (.not. spin) then
           if (fresh_incr) then
-          ! apply EnKF increments 
+!sqz 2020 -- save bkgd cat_progn for diagaup_forecast
```
gmao-rreichle, line 1879: not needed, see comment above

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1793,6 +1959,31 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
...
+             call get_diagaup('analysis', NUM_ENSEMBLE, N_catl, cat_param, cat_progn, &
+                                diag_aup_ensavg)
+
+              if(associated(SFSM_ana)) SFSM_ana(:) = diag_aup_ensavg(:)%srfexc  ! borrowed names
+              if(associated(RZSM_ana)) RZSM_ana(:) = diag_aup_ensavg(:)%rzexc   !see get_diagaup
+              if(associated(PFSM_ana)) PFSM_ana(:) = diag_aup_ensavg(:)%catdef  ! 
+              if(associated(TPSURF_ana)) TPSURF_ana(:) = diag_aup_ensavg(:)%tc1
+              if(associated(TSOIL1_ana)) TSOIL1_ana(:) = diag_aup_ensavg(:)%tc2
+!
+           call get_diagaup('forecast', NUM_ENSEMBLE, N_catl, cat_param, cat_progn_fcst, &
```
gmao-rreichle, line 1978: the 2nd part of this new block dealing with *_fcst should not be needed

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -2932,6 +2935,114 @@ end subroutine write_smapL4SMaup
 
   ! -----------------------------------------------------------------
 
+!sqz 2020 -----
+  ! **********************************************************
+
+ subroutine get_diagaup (option, N_ens, N_catl, cat_param, cat_progn, diagaup)
```
gmao-rreichle, line 2941: I'm not sure why this subroutine needs to distinguish between the "forecast" and the "analysis".
In a nutshell, we should only need a wrapper for catch_calc_soilmoist(), catch_calc_tsurf(), and catch_calc_tp().
If we send the updated variables (ie, cat_progn after the increments have been applied) into the subroutine, we should then get the desired *_ana diagnostics back.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -2932,6 +2935,114 @@ end subroutine write_smapL4SMaup
...
+
+          call catch_calc_tp( N_catl, cat_param%poros,                            &
+               catprogn2ghtcnt(N_catl,cat_progn(:,n_e)), tp )
+
+          tp1(:,n_e) = tp(1,:) + MAPL_TICE
+
+       end do  !n_e N_ens 
+
+       ! compute ensemble mean values and std-dev
+       call row_std( N_catl, N_ens, sfmc,  tile_std_l(:,1), tile_mean_l(:,1) )
```
gmao-rreichle, line 3010: The LDASsa "aup" files include not only the (ensemble-mean) FCST and ANA variables but also the ensemble std-dev.  The above edits ignore the possibility of adding ens-std-dev values into the "lndfcstana" file, which we will probably want later on...
For now perhaps just keep it to the ens mean, but keep in mind that later on we'll probably want to add the ens-std-dev as well.  We would then have to decide again whether to do the ens stats here or in the Ens Grid Comp.

saraqzhang:
we certainly can skip "fcst" part and only get "ana" set in this "aup" collection.   In that way we rely on the inst  03h lnd collection for "fcst" values.  It probably is desirable to have sort of instruction to users in generating History.rc that if the analysis "aup" collection is selected, then inst 03h lnd should be selected as well.

gmao-rreichle:
@saraqzhang : I think you misunderstood, probably because I had a typo in one of my comments, sorry.  The following lines would be in the _same_ collection ("inst3_1d_lndfcstana_Nt") in HISTORY.rc:

'WCSF' , 'ENSAVG' , 'SFMC_FCST'
'WCSF_ANA' , 'LANDASSIM' , 'SFMC_ANA'

So if the user selects the "inst3_1d_lndfcstana_Nt", both the FCST and ANA variables will be included in the same file.  The trick is that the FCST variables can just be pulled from the Ens Grid Comp.  They don't need to be recomputed in the LandAssim Grid Comp.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
saraqzhang: currently we borrowed different variable names in the "cat_progn" type  to hold "fcst" (5) and "ana" (5),  "forecast" or "analysis" tells the subroutine which variable names ( in the type)  to be put into for locally calculated tsurf etc.  the "type" is the one to communicate between local routine and the assimGC.   
if we skip "fcst" in the export, we do not need "forecast" or "analysis" here.

saraqzhang:
yes that will do.  it will be a ldas analysis-specific inst collection,  not the general inst lnd collection that is more GCM general collection.

saraqzhang:
@gmao-rreichle :  I made modifications ( lndfcstana collection)  according to the discussion last week and pushed to the branch Feature/diagAandB,  the modification passed test run.

gmao-rreichle:
addressed with pull request #240 (after starting a different branch and pull request)

PR #213, opened 2020-05-07
remove unnecessary comments in execinput file

weiyuan-jiang:


src/Applications/LDAS_App/GEOSldas_LDAS.rc
```diff
@@ -53,13 +53,6 @@ LSM_CHOICE: 1
 # WHITE_FILE: '' 
 
 
-# ---- Surface meteorological forcing: Time step
```
gmao-rreichle, line 56: I don't think we can just remove this.  While MERRA-2 and GEOS FP forcing use the default (3600s, or 1 hour), other legacy forcing data have different values (e.g., 3 hours). 
Is there a reason why this was removed?

weiyuan-jiang: This is in the "required" field

weiyuan-jiang: Maybe we can move up the comments

gmao-rreichle: Sorry, I didn't catch that this is already put into the *.exec through "ldas_setup" and thus ended up there twice.  Nice catch.  I will edit the comment in "ldas_setup".  Stay tuned.

Issue #139, opened 2020-02-27
Need "increments" output in units of soil moisture, temperature, etc


gmao-rreichle:
The existing "catch_progn_incr" collection writes out soil moisture increments in terms of Catchment model prognostics variables.  Without knowing the model forecast (or the analysis) in the same variables, the increments cannot be converted into units of soil moisture, and even if we knew the model forecast it would not be simple because we do not have an offline tool to convert Catchment prognostics into soil moisture units.
The "aup" output for the SMAP system does provide the model forecast and analysis in units of soil moisture, but this output type is limited to the assimilation of SMAP (or SMOS) obs and when running on the SMAP 9-km EASE grid tiles.
For more general output of increments, we need to add a few key model "forecast" (and perhaps also "analysis") variables as "exports" in the land assim Grid Comp.  Then HISTORY can be used to write "forecast" and "analysis" variables (incl in soil moisture units), from which the increments can be derived in post-processing.
@weiyuan-jiang and @gmao-rreichle should discuss details during weekly tag-up meeting.

saraqzhang:
@gmao-rreichle :  I just put in a pull request  #212 for this issue.  I don't know if I should revert the pull request to draft. let me know.

gmao-rreichle:
addressed in pull request #240

PR #240, opened 2020-06-05
Forecast and analysis output of model diagnostics via HISTORY

gmao-rreichle:
addresses issue #139 
replaces pull request #212 

I created a revised branch starting from branch feature/diagAandB by @saraqzhang 

Successfully passed nightly tests.

Verified "lndfcstana" output vs "aup" output on EASE grid for LDAS_GLOBAL/assim test case.

Differences between "lndfcstana" and "aup" fields are greater than roundoff but acceptable for science.  Differences likely stem from different paths for the calculation of the ensemble average diagnostics.  Further modifications to reduce the differences is left for future work.

Also left for future work is modification of lenkf.j.template so that "lndfcstana" fields are moved into ./ana instead of ./cat (TBD).
 
@saraqzhang : Please take a look and test for assimilation on cube-sphere tiles.


saraqzhang:
@gmao-rreichle , I pulled  the  branch and ran cs assim. test. results are verified by checking A-F ( consistent with increment location and amplitude ) ,  and F-inst3_lnd  ( zero diff since the progn pert. is moved up ) .  -Sara

PR #276, opened 2020-07-15
Feature/add landpert rst file

weiyuan-jiang:
Add landpert_restart_file parameter. If it does not exist, the values will be set to zero


gmao-rreichle:
@weiyuan-jiang confirmed that MAPL initializes the LANDPERT seed values to zero.
It is therefore ok to have the LANDPERT restart file line in LDAS.rc even when the restart file is not available initially.

PR #277, opened 2020-07-15
bug fix: convert lat/lon (centerY/centerX) for EASE tiles to radians before use in ESMF

gmao-rreichle:


gmao-rreichle:
centerX and centerY for EASE grid were not used yet, so should be 0-diff-trivial

gmao-rreichle:
full nightly test completed successfully

PR #295, opened 2020-08-18
Feature/use_hash_indg

weiyuan-jiang:
Keep original i_indg and j_indg  and use has_indg and hash_j_indg for the computation

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
@@ -401,18 +401,20 @@ contains
        ! from lat/lon to tiles.  This needs to be done:
        ! - by root process (because of call to read_obs() in collect_obs())
        ! - by all processes if FOV>~0 ("tile_num_in_circle" needed in get_obs_pred())
-       if ( (root_proc)                                        .or.            &
+       if ( (root_proc)                                        .or.              &
             (any(obs_param(1:N_obs_param)%FOV>FOV_threshold))        )  then
 
           allocate(N_tile_in_cell_ij_f(tile_grid_f%N_lon,tile_grid_f%N_lat))
 
           ! first call: count how many tiles are in each tile_grid_f cell
-          call get_number_of_tiles_in_cell_ij( N_catf, tile_coord_f, tile_grid_f,    &
-               N_tile_in_cell_ij_f )
+          call get_number_of_tiles_in_cell_ij( N_catf,                           &
+               tile_coord_f%hash_i_indg, tile_coord_f%hash_j_indg,               &
+               tile_grid_f, N_tile_in_cell_ij_f )
           ! second call: find out which tiles are in each tile_grid_f cell
 
-          call get_tile_num_in_cell_ij( N_catf, tile_coord_f, tile_grid_f,    &
-               maxval(N_tile_in_cell_ij_f), tile_num_in_cell_ij_f )
+          call get_tile_num_in_cell_ij( N_catf,                                  &
+               tile_coord_f%hash_i_indg, tile_coord_f%hash_i_indg,               &
+               tile_grid_f, maxval(N_tile_in_cell_ij_f), tile_num_in_cell_ij_f )
        end if
 
        ! *********************************************************************
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
@@ -2738,10 +2738,10 @@ contains
                 data_h_9km_tile_8 = real(nodata_generic,kind(0.0D0)) ! init (not in grid2tile!)
                 data_v_9km_tile_8 = real(nodata_generic,kind(0.0D0)) ! init (not in grid2tile!)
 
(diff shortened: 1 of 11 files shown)


gmao-rreichle:
replaces pull request #270

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -2738,10 +2738,10 @@ subroutine write_smapL4SMaup( option, date_time, work_path, exp_id, N_ens,    &
                 data_h_9km_tile_8 = real(nodata_generic,kind(0.0D0)) ! init (not in grid2tile!)
                 data_v_9km_tile_8 = real(nodata_generic,kind(0.0D0)) ! init (not in grid2tile!)
 
-                call grid2tile( tile_grid_g, N_catf, tile_coord_f%i_indg, tile_coord_f%j_indg, data_h_9km_grid_8, &
+                call grid2tile( tile_grid_g, N_catf, tile_coord_f%hash_i_indg, tile_coord_f%hash_j_indg, data_h_9km_grid_8, &
```
gmao-rreichle, line 2741: I'm pretty sure this should be %i_indg, not %hash_i_indg.  Here, we are simply going back and forth between the tile space and the grid that is associated with this tile space.  Because this subroutine only works for the EASE grid, and because %i_indg and %hash_i_indg are identical in this case, the results are going to be the same.  But we still need to use the appropriate field of the tile_coord_f structure.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -2738,10 +2738,10 @@ subroutine write_smapL4SMaup( option, date_time, work_path, exp_id, N_ens,    &
...
                      data_h_9km_tile_8 )
 
-                call grid2tile( tile_grid_g, N_catf, tile_coord_f%i_indg, tile_coord_f%j_indg, data_v_9km_grid_8, &
+                call grid2tile( tile_grid_g, N_catf, tile_coord_f%hash_i_indg, tile_coord_f%hash_j_indg, data_v_9km_grid_8, &
```
gmao-rreichle, line 2744: Same comment as above.

weiyuan-jiang: I feel we should not cheery-picking but be consistent through out the codes: hash_x_indg  is used for computation and x_indg is used to keep the original record whenever we can. The only place we should use X_indg probably is here https://github.com/GEOS-ESM/GEOSldas/blob/b674b877cac911106b55b5f3389196fa9cc9d1c3/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90#L4674-L4679

gmao-rreichle: This isn't cherry-picking.  For a grid2tile operation, we need to use the i,j indices that are associated with grid that is involved in the operation.  Here, the grid is the original grid associated with the tile space, not the "hash" grid used in the perturbations and to support the EnKF analysis.   I know this is a bit subtle because it looks like it's part of the EnKF analysis, but it's really just a complicated special case to __output__ (1d array) tile-space fields in gridded (2d array) format (taking advantage of the fact that by construction the EASE-based tile space has at most one tile in each grid cell.) 
Again, the grid involved here is NOT the hash grid that is used for perturbations or to support the EnKF analysis.

weiyuan-jiang: It seems it is just part of SMAP writing. Changed

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -4672,10 +4672,10 @@ subroutine GEOS_openfile(FileOpenedHash, fname_full, fid, tile_coord, m_hinterp,
        ! compute indices for nearest neighbor interpolation from GEOSgcm grid
        ! to tile space
           if( isCubed ) then ! cs grid
-             ! i_indg and j_indg are changed to LatLon grid
+             ! cube-sphere grid of forcing data must match cube-sphere grid associated with tile space
```
gmao-rreichle, line 4675: I think we double-check this somewhere in ldas_setup or preprocessing.  Is that correct?

weiyuan-jiang: No, it is not correct.  see my comment below

gmao-rreichle: After checking with @weiyuan-jiang , it turns out we do not verify that the cube-sphere grids in question match.   This is left for later.

gmao-rreichle:
The previous comment (by @weiyuan-jiang ) applies to pre-processing.
After discussion, it was agreed that the "%hash_[x]_indg" fields should not be needed in pre-processing and should be initialized to no-data.  
To resolve this, the interface to "get_tile_grid()" needs to be changed to accommodate using the subroutine in two different ways.  One possibility is to add an optional argument that tells the subroutine whether to use %hash_[x]_indg or %[x]_indg.
Another possibility is to pass in the required integer and real arrays instead of the tile_coord structure.
More elegant options may exist.

weiyuan-jiang:
The program crashes. Need to figure out why

PR #310, opened 2020-09-14
add message to "assert" statements

weiyuan-jiang:
When assert fails, it would report more accurate information


gmao-rreichle:
@weiyuan-jiang : Thanks for tackling this.  I added one commit with minor changes and will approve the pull request so that you can proceed with merging if you agree with my changes.  But don't hesitate to iterate if there's anything that looks wrong to you. 
IMPORTANT: I did NOT test the branch after my commit.  If you agree with my changes, it's probably best to re-test the branch before merging to be safe.

gmao-rreichle (approved):

PR #312, opened 2020-09-21
change NUM_LDAS_ENSEMBLE in one place

weiyuan-jiang:
It is more convenient and consistent to change NUM_LDAS_ENSEMBLE in just one place while debugging or testing

### DIFF
--- a/src/Applications/LDAS_App/lenkf.j.template
+++ b/src/Applications/LDAS_App/lenkf.j.template
@@ -74,7 +74,7 @@ endif
 
 if (! -e $SCRDIR            ) mkdir -p $SCRDIR
 
-set NENS = `grep NUM_LDAS_ENSEMBLE   $HOMDIR/LDAS.rc | cut -d':' -f2`
+set NENS = `grep NUM_LDAS_ENSEMBLE:  $HOMDIR/LDAS.rc | cut -d':' -f2`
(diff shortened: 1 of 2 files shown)


src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -75,6 +74,7 @@ endif
 
 if (! -e $SCRDIR            ) mkdir -p $SCRDIR
 
+set NENS = `grep NUM_LDAS_ENSEMBLE   $HOMDIR/LDAS.rc | cut -d':' -f2`
```
gmao-rreichle, line 77: Should there be a colon after "grep   NUM_LDAS_ENSEMBLE"?  There is a colon after the grep for END_DATE and NUM_SGMT.

weiyuan-jiang: It is not necessary. The grep gets the whole line and cut by ":".  But we can add it if we want the syntaxs to be consistent.

biljanaorescanin (approved):
This was during GEOSldas tagup meeting

gmao-rreichle:
When this is merged, please add something like the following "squashed" commit statement:

"Determine NUM_LDAS_ENSEMBLE entry in lenkf.j from LDAS.rc, replacing the independent, ldas_setup-generated specification in lenkf.j"

weiyuan-jiang:
will be addressed by #316

PR #313, opened 2020-09-22
fix for max(xhalo, ...)

weiyuan-jiang:
The intrinsic function "max" needs the arguments with the same shape ( except the scalar).

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
@@ -18,7 +18,8 @@ module clsm_ensupd_upd_routines
        logit,                                     &
        logunit,                                   &
        nodata_generic,                            &
-       nodata_tol_generic
+       nodata_tol_generic,                        &
+       LDAS_is_nodata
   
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
I've been staring at the revisions for a while... I'm nervous that it's rather risky to assig a distance of 0. in subroutine `dist_km2deg()` when the input is a no-data-value.  This works for the use case in question, where we take a `max()` of what comes back from `dist_km2deg()`.  But what if `dist_km2deg()` is used in a way that is not followed by a `max()` operation?

From the description above I'm guessing that the offending line is the following: 

` xhalo = max( xhalo, tmprx(1:kk) ) `

Specifically, I understand that the problem in the statement above is that `xhalo` is a scalar and `tmprx(1:kk)` is a vector.  

Could we just replace the above line with the following:

`xhalo = max( xhalo,  maxval( tmprx(1:kk) ) )` 

Maybe that's a simpler way to achieve the same result, without the risky modification to `dist_km2deg()`.

weiyuan-jiang:
Xhalo is a vector. That is the source of the problem. We can assign nodata to dist_km2deg if the input is nodata.

weiyuan-jiang:
I am wondering if the dist_km2deg works for lat-lon grid

weiyuan-jiang:
closed.  it is addressed by #316

PR #316, opened 2020-09-24
Hotfix/no_tiles_in process

weiyuan-jiang:
1) fix the situation when there is no tiles in a process
2) Determine NUM_LDAS_ENSEMBLE entry in lenkf.j from LDAS.rc, replacing the independent, ldas_setup-generated specification in lenkf.j
3) use the relative link path. When the whole directory is moved (like CURRENT to BASELINE) the link would be still valid.

### DIFF
--- a/src/Applications/LDAS_App/ldas_setup
+++ b/src/Applications/LDAS_App/ldas_setup
@@ -1278,8 +1278,6 @@ class LDASsetup:
                         fout.write(line.replace('MY_EXPID',self.rqdExeInp['EXP_ID']))
                     elif 'MY_EXPDOMAIN' in line :
                         fout.write(line.replace('MY_EXPDOMAIN',self.rqdExeInp['EXP_DOMAIN']))
-                    elif 'MY_ENSEMBLE' in line :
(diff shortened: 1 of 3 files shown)


gmao-rreichle:
@weiyuan-jiang : I fixed the use of the optional argument "obsbias_ok_tmp" and added a few comments, see commit [`29deeed`](https://github.com/GEOS-ESM/GEOSldas/pull/316/commits/29deeedf7ca6de206360c5ef146e7da466d3b38b)
Did you already test this branch?  (My commit should not change the test result.)

cc: @biljanaorescanin

weiyuan-jiang:
good catch. I have test this branch and it seems zero-diff

gmao-rreichle (approved):
Approved.  I understand the comment by @weiyuan-jiang to mean that the branch is 0-diff vs develop in the current test suite.  What remains to be seen is if the branch is 0-diff between the (old) 1d decomposition (w/ optimization) and the new 2d decomposition functionality. 
In any case, I'm ok with merging the branch at this time because it is 0-diff in the current test suite and because it fixes other things. 
But we may need additional modifications to achieve 0-diff between the 1d and 2d decompositions.
cc: @biljanaorescanin

PR #317, opened 2020-10-06
change MAPL_Base to MAPL

weiyuan-jiang:
In the near future, MAPL_Base is not supported anymore


tclune (approved):
Sorry about that one.  Hopefully this is now a stable way to depend on MAPL.

Issue #114, opened 2020-02-14
soil (and snow?) temperatures in units of Kelvin throughout ; do not write TSURF into "offline" catch restarts


gmao-rreichle:
Work with the GCM group to convert all soil (and snow?) temperature EXPORT variables outside of catchment.F90 into units of Kelvin.
Currently, units are changed only at the level of the Surface Grid Comp or in the GEOSldas Ensemble GridComp, which is confusing because some EXPORTS are in Celsius and others in Kelvin.

smahanam:
@weiyuan-jiang are you going to do this? Let me know if you want me to do it.

gmao-rreichle:
@smahanam : Because this change isn't guaranteed to be zero-diff, it's probably better to wait.  I first need to talk with Scott to figure out a good opportunity to get this adopted into the GCM.

Another thing: When working on the GCM Grid Comp, we should also change the code such that TSURF is no longer written into "offline" catch restart files.  This can be done by changing 

  RESTART            = MAPL_RestartRequired

to 

  RESTART            = RESTART

in the InternalSpecs definition of TSURF.

smahanam:
TP1, TP2 ... are model diagnostics. This modification should pass AGCM zero-diff.
Perhaps, GEOSldas nightly test will fail, because the nightly tests (I think) diff all files.

gmao-rreichle:
@smahanam : It's true that this is zero-diff in terms of restarts.  I'm not sure what exactly is tested for the AGCM.  I also don't know if the AGCM team would accept a change that impacts diagnostics as easily as a truly zero-diff change.  I hope to catch up with Scott next week and come up with a plan for getting this and other current, land-related "issues" into the GCM.

PR #329, opened 2020-11-04
Soil temperatures (TP's) in units of Kelvin throughout

biljanaorescanin:


gmao-rreichle:
@biljanaorescanin, here are my 2c:

1.) The changes in your branch are probably too extensive.  The original objective of #114 was to "convert all soil (and snow?) temperature **EXPORT** variables **outside of catchment.F90** into units of Kelvin."  We should probably consider the model diagnostic subroutines (e.g., catch_calc_tp()) to be part of catchment().  For one thing, the diagnostics subroutines are also used in the ./mk_restarts, which would presumably require further edits.   Second, the snow temperatures are still a mix of Celsius and Kelvin.  Changing snow temperatures to Kelvin throughout would require us to work with the landice group.  So it's probably best to stick with the original plan and only change the units for "export" variables.  That is, soil temperatures remain in Celsius for all of the scientific computations associated with catchment() and are only converted to Kelvin when the variables become MAPL exports.  
Restarting from `develop`, I created two new branches, one for GEOSldas and one for GEOSgcm_GridComp, with the same name: `feature/rreichle_TP_C_to_K`.
I successfully tested the GEOSldas branch (except for the CatchCN test, see below, and for Intel only)

2.) My new GEOSgcm_GridComp branch also removes TSURF from the list of variables that are written into offline Catch and CatchCN restart files (per #114).  This also seems to work as intended, but (i) it results in an obvious "failure" of the comparison of the restart files and (ii) it may break the code in ./mk_restarts or LDAS_App/mk_GEOSldasRestarts.F90, which is *not* exercised in the "nightly" tests.  Maybe @weiyuan-jiang can comment on what changes might  be needed, if any.

3.) I ran my tests only for Catch because the CatchCN test still needs to be updated.  Before we can finalize this pull request, we need to fix the CatchCN test and then test this branch.  I also didn't do the GNU tests yet, or the AGCM tests.

Please take a look and review my branch (see above).  If it makes sense to you, please update the branch associated with this pull request accordingly. 

cc: @gmao-jkolassa @rdkoster

biljanaorescanin:
@gmao-rreichle Yes, all make sense and look good. The introduction of the optional tp1_in_Kelvin argument into catch2mwRTM_vars subroutine is a clean way to solve for the differences in the input units from various calls. I'll review the branch and proceed with updates once we fix the CatchCN tests and reiterate with @weiyuan-jiang to see about potential issues listed under 2.) 
cc: @gmao-jkolassa @rdkoster

biljanaorescanin:
Addressed with PR #340

PR #340, opened 2020-12-02
Soil temperatures (TP's) EXPORT variables in units of Kelvin throughout

biljanaorescanin:
Currently, units are changed only at the level of the Surface Grid Comp or in the GEOSldas Ensemble GridComp, see Issue #114.  This is confusing because some EXPORTS are in Celsius and others are in Kelvin.
A first implementation was discussed in PR #329, which contains more discussion (the associated branch was discarded and is superseded by this PR).
The matching GEOSgcm_GridComp PR is [#387](https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/387).  The related GEOSgcm_GridComp branch also removes TSURF from the list of variables that are written into offline Catch and CatchCN restart files.


biljanaorescanin:
@weiyuan-jiang once you confirm that what is done for TSURF is ok, I will add it to description of what is done in this PR. 
I can add Nightly tests results summary as well so we have all at one place.

gmao-rreichle (approved):

biljanaorescanin:
Tests summary as expected, non zero diff changes: 
Output NC4 files: All model runs fail the comparisons to the baseline for LND collection for variables TP1-TP6 due to the conversion of soil temperature units (values) from C to K. 
All assimilation runs fail in GPH collection for the same reason - the conversion from C to K causes a round off difference for soil_temp_layer1- soil_temp_layer6. Round off difference also in collection inst3_2d_lndfcstana for TSOIL1_FCST variable.
Fill Value: TPSNOW, TUNST, TSAT and TWLT are different in history for information on the fill value (1.e+15f)
Output RST: catch.internal_rst file fails comparison due to removal of TSURF from the list of variables that are written into offline Catch and CatchCN restart files.
@gmao-rreichle @weiyuan-jiang

gmao-rreichle (approved):

PR #341, opened 2020-12-09
Feature/wjiang/output_incr

weiyuan-jiang:
add a GridComp to export catch_progn_incr for individual ensemble member.


tclune (dismissed):
CMake changes ok.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -15,7 +15,7 @@ module GEOS_LandAssimGridCompMod
   use ESMF
```
gmao-rreichle, line 15: The comments ("DESCRIPTION") before this line should be edited because LandAssimGC now does have Children.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -15,7 +15,7 @@ module GEOS_LandAssimGridCompMod
...
   use MAPL_Mod
   use ESMF_CFIOMOD,              only: ESMF_CFIOstrTemplate
-  
+  use GEOS_OutputIncrGridCompMod,only: OutputIncrSetServices=>SetServices
```
gmao-rreichle, line 18: Re. the name of the new GridComp:
1) I think it would be better to use "Export" instead of "Output".  The term "export" suggests that this is about MAPL.  The term "output" is used in the legacy LDASsa subroutines that bypass MAPL.
2) I'm not sure if we should include "Land" or "Ldas" in the GC name.  This GC is in a sub-dir of the LandAssim GC, so it may not be necessary.
So maybe change "GEOS_OutputIncrGridCompMod" to "GEOS_ExportIncrGridCompMod" (with similar changes for the corresponding F90 file and dir name)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -145,6 +145,9 @@ subroutine SetServices ( GC, RC )
     type(MAPL_MetaComp), pointer :: MAPL=>null()
     type(ESMF_Config)            :: CF
     character(len=ESMF_MAXSTR)   :: LAND_ASSIM_STR, mwRTM_file
+    character(len=ESMF_MAXSTR)   :: id_string,childname, fmt_str
+    integer                      :: i, ens_id_width, FIRST_ENS_ID, NUM_ENSEMBLE
+    integer, allocatable, dimension(:) :: ens_id, output_incr
```
gmao-rreichle, line 150: If I'm reading this correctly, there's a variable name "output_incr" and also a subroutine name "output_incr" and, on top of that, the children GCs are labeled OUTPUT_INCR[nnnn]).  
I guess the variable and subroutine are separate things and Fortran sorts out from the context which is which.  
Would  it be better practice to use a different name for, say, the subroutine?  See comment below.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -2338,7 +2278,129 @@ subroutine OUTPUT_SMAPL4SMLMC(gc, import, export, clock, rc)
      _RETURN(_SUCCESS)
 
   end subroutine OUTPUT_SMAPL4SMLMC 
- 
+
+  subroutine OUTPUT_INCR( cat_progn_incr, export,rc)
```
gmao-rreichle, line 2282: If I'm reading this correctly, the subroutine output_incr() essentially copies data from the legacy LDASsa "cat_progn" structure into the MAPL export variables. 
I suggest renaming the subroutine (also to avoid confusion with the variable name "output_incr" above).
Maybe output_incr() could be rename to
   export_incr()
or 
   catprogn2export_incr()
or something else along those lines (avoiding the term "output" -- see comment above).

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOSoutputincr_GridComp/GEOS_OutputIncrGridComp.F90
```diff
@@ -0,0 +1,391 @@
...
+         gc,                                                                    &
+         ESMF_METHOD_FINALIZE,                                                  &
+         Finalize,                                                              &
+         rc=status                                                              &
+         )
+    _VERIFY(status)
+        
+  ! Exports for Catchment prognostics increments
+
+  call MAPL_AddExportSpec(GC                                                      ,&
```
gmao-rreichle, line 90: This list of calls to MAPL_AddExportSpec() for the increments is also present verbatim in GEOS_LandAssimGridComp.F90 (beginning ~line 227).
I'm guessing this child GC F90 file is the right place to define the list (w/ LONG_NAME and UNITS).
But do we need to repeat this list in GEOS_LandAssimGridComp.F90?
Maybe we can use the syntax that is used in GEOS_LandGridComp.F90 to bubble up the exports from Catch to Land?  That is, can we replace the entries here with something like
`   call MAPL_AddExportSpec(GC, SHORT_NAME = 'TCFSAT_INCR', CHILD_ID = CATCH(1), RC=STATUS); VERIFY_(STATUS)
`
etc?
This might not work because here we're dealing with ensemble members, but maybe it does.  If so, it would be good to have to define LONGNAME and UNITS in only one place.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: I changed it to GEOS_ExportCatchIncrGridCompMod. It seems more clear now

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: I change those names to distinguish them.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: Since it is private to the module, the subroutine name Export_INCR should be enough

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOSoutputincr_GridComp/GEOS_OutputIncrGridComp.F90
weiyuan-jiang: I don't think we can do that. The export variables in LandAssim  are different from those in  ExportCatchIncr even though the names are the same (Sara told me the name should be the same). If there are no collections in HISTORY.rc to export the incr of individual member , the export won't be allocated and initialized.

gmao-rreichle (dismissed):
Looks good, thanks!   I committed a couple of minimal edits to comments ("DESCRIPTION" of LandAssim and ExportCatchIncr GridComps).
@weiyuan-jiang : Once you're happy with the PR and @saraqzhang confirms that everything works as intended, please forward to @biljanaorescanin for merging.

weiyuan-jiang:
@saraqzhang . The grid comp names are changed from OUTPUT_INCRxxxx to CATCHINCRxxxx.  The xxxx is the ensemble ID. You need to update your HISTORY.rc to run the test again.

weiyuan-jiang:
Two scripts are are added to LDAS_App/util/.  One is used to calculate forcing ensemble average, the other is used to generate HISTORY.rc for individual catch incr.

saraqzhang:
@weiyuan-jiang  @gmao-rreichle   I pulled the updates  and ran test ( GC name change, two util scripts) , all worked as expected.

weiyuan-jiang:
@saraqzhang , The script that calculates the ensemble average forcing requires all the files in the folders have same names and numbers. That makes sense.  You can change a file name , add a file in a folder to test it. That can build the confidence that the script really produces what we want.

saraqzhang:
@weiyuan-jiang @gmao-rreichle ,  about the scripts: 
(1)  average_ensemble_forcing.py:   in my test I added  *lfo*  in the line    fs = sorted... 
since in adas ensdiag/mem***  there will be other diag files as well, we only do ens avg on lfo.  
I also tested the cases of different file name, additional file in one mem/,    py script generated assertion error and no avg file is created for the case. 
(2)  generate_catchincr_hist.py :   catch_progn_incre***_descr:  probably we should remove  '... Ensemble-Average...'

weiyuan-jiang:
The two issues are addressed @saraqzhang

biljanaorescanin (approved):

biljanaorescanin:
Nightly tests summary: All tests pass. 
@gmao-rreichle @weiyuan-jiang @saraqzhang

src/Applications/LDAS_App/util/average_ensemble_forcing.py
```diff
@@ -0,0 +1,42 @@
...
+
+def averaging_forcing(in_path, out_path, nens):
+   """ The ensemble number will be appended to in_path
+       it starts from 001
+       out_path will be created if it does not exists """ 
+   if not os.path.exists(out_path):
+      os.makedirs(out_path) 
+   files_list=[]
+   for i in range(1,nens+1):
+      sfx = f'{i:03}'
```
gmao-rreichle, line 18: @saraqzhang :  I'm guessing that this line means we use 3 digits for the ensemble ID string.  Is this the practice in the ADAS?  In the LDAS, we've typically been using 4 digits, as in "ens0018", for example.  Unless there's some ADAS-related reason, it'd probably be better to use 4 digits for consistency with the rest of the LDAS.  I might be completely misinterpreting this piece of code, though.  Apologies if that's the case.

saraqzhang: @gmao-rreichle:  @weiyuan-jiang  ADAS atmens use 3 digits for member ID. this script is to access all member subdir in ADAS atmens to calculate ens average of the forcing.  Should we put some comments in the script specifying the script is for processing hybrid ADAS outputs.   If we wanted to make the script more general, we need to make user input argv for the digits, and the file name etc.

gmao-rreichle: @saraqzhang : Thanks for confirming that 3 digits is ADAS practice.  In this case it's fine to leave things as they are.  I suppose at a later time we should make these new py scripts more flexible.  As you put together the LADAS, you'll find out which variables would be good to have as inputs to the script, or at least have them set at the top of the script (rather than being buried in the code below).  For now let's move ahead with the PR.
cc: @weiyuan-jiang @biljanaorescanin

mathomp4 (approved):
Approving for CMake

PR #346, opened 2021-01-05
Updates to GEOSldas components.yaml

mathomp4:
This is a zero-diff, yet scary-looking, update to `components.yaml` which brings it in line (or slightly ahead, pending PRs) of GEOSgcm:

1. Updates to ESMA_cmake v3.3.5 and ecbuild geos/v1.0.6
    * These are mainly feature updates that support a new "Aggressive" build type which turns on all the flags for speed (at the cost of MPI layout transparency and maybe stability, never tested with LDAS)
2. Updates to ESMA_env v3.1.3
    * This is mainly an update to Baselibs 6.0.27 which is mainly fixes to the Baselibs build system as well as a few minor library updates (yaFyaml, cURL, NCO)
3. Converts the file to use "relative" paths to sub repos. 
    * This style is based off of the "relative" paths used in git submodules. This is mainly useful for things like CI where you can't assume someone has SSH access to GEOS-ESM. Thus, now, if someone does a `mepo clone https://github.com/GEOS-ESM/GEOSldas.git`, the subrepos will *also* be cloned with https. Likewise `mepo clone git@github.com:GEOS-ESM/GEOSldas.git` will use ssh URLs inside.
4. Adds the "fixture" entry for GEOSldas which lets mepo see the fixture itself!

Of course, this requires @biljanaorescanin for testing, but I can't see any issues...

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -1,36 +1,40 @@
+GEOSldas:
+  fixture: true
+  develop: develop
+
 env:
   local: ./@env
-  remote: git@github.com:GEOS-ESM/ESMA_env.git
-  tag: v3.1.1
+  remote: ../ESMA_env.git
+  tag: v3.1.3
   develop: main
 
 cmake:
   local: ./@cmake
-  remote: git@github.com:GEOS-ESM/ESMA_cmake.git
-  tag: v3.3.1
+  remote: ../ESMA_cmake.git
+  tag: v3.3.5
   develop: develop
 
 ecbuild:
   local: ./@cmake/@ecbuild
-  remote: git@github.com:GEOS-ESM/ecbuild.git
-  tag: geos/v1.0.5
+  remote: ../ecbuild.git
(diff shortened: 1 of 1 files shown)


mathomp4:
Moving to draft. I'm seeing something weird locally...

biljanaorescanin:
Since we figured out what quirk was I will run the tests now. @mathomp4 I'll keep you updated if all goes well we can upgraded it to full PR.

mathomp4:
Yes. Making ready for review. This PR exposed a bug in `mepo` itself. Or, rather, a quirk of `git` that `mepo` was exposing. I don't know. Fixed now.

biljanaorescanin:
I can confirm, since all tests are done, it is a zero diff change. 
@mathomp4 @gmao-rreichle @weiyuan-jiang

biljanaorescanin (approved):

PR #347, opened 2021-01-07
Update LADAS_COUPLING options to support coupling with ensemble component of ADAS

saraqzhang:
add options for coupling with enadas:  facilitates coupling of LDAS with ensemble component of ADAS (using ensemble-average of surface met forcing)


saraqzhang:
@weiyuan-jiang @gmao-rreichle: minor changes for the option of coupling with enadas ( process ensemble forcing) .

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -68,6 +68,18 @@ else
...
 endif
 
+#######################################################
+#  if LADAS_COUPLING ==2, do atmens forcing processing 
+###################################################### 
+   if ( $LADAS_COUPLING == 2) then
+     cd $HOMDIR
+     module load  python/GEOSpyD/Ana2019.03_py3.7
+     module  load nco/4.8.1
+     python  average_ensemble_forcing.py
```
gmao-rreichle, line 78: Doesn't this script have to be added to the repository?

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -68,6 +68,18 @@ else
    exit
 endif
 
+#######################################################
+#  if LADAS_COUPLING ==2, do atmens forcing processing 
+###################################################### 
+   if ( $LADAS_COUPLING == 2) then
+     cd $HOMDIR
+     module load  python/GEOSpyD/Ana2019.03_py3.7
+     module  load nco/4.8.1
```
gmao-rreichle, line 77: @weiyuan-jiang :
We also load nco in process_rst.csh (where we can probably clean up the if block that was needed for the SLES11-to-SLES12 transition period).
Is there a better way to loading these modules through the environment? 
Perhaps nco isn't by default in the g5 environment, but how about python?

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -68,6 +68,18 @@ else
    exit
 endif
 
+#######################################################
+#  if LADAS_COUPLING ==2, do atmens forcing processing 
+###################################################### 
+   if ( $LADAS_COUPLING == 2) then
```
gmao-rreichle, line 74: @saraqzhang : 
I'm slightly surprised this if block occurs at the very beginning of the job script.  Without seeing the python script, it's hard to tell which exact forcing data are ensemble-averaged.  
I assume that *all* of the forcing data that are needed for the 6-hour (?) interval are already staged before the LDAS simulation starts?   
Put differently, I'm wondering just how much of the current ADAS configuration we're hardwiring into the LDAS portion of the LADAS.

src/Applications/LDAS_App/lenkf.j.template
saraqzhang: it was added in app/util last month  (the the same time with the merge for ensemble increments).

src/Applications/LDAS_App/lenkf.j.template
saraqzhang: @gmao-rreichle 
yes all of the forcing for current 6h segment are staged at enadas ( generated by the ensemble predictor segment in the last cycle of enadas).  they are available before the enadas process start the current cycle. 
 enadas does not produce ensmean for diag ( forcing included). so we do forcing averaging at ldas side when the run is coupled to enadas.    The enadas configuration such as stage dir path and ens sub dir structure are hardwired and limited  in the python code only.  Otherwise not hardwired but  specified at setup time (  ens size matching that of enadas, forcing path, forcing file tag etc.) .

src/Applications/LDAS_App/lenkf.j.template
weiyuan-jiang: I think at this point, nco is already loaded, so it is not necessary here.

weiyuan-jiang:
@saraqzhang Can you try this branch feature/wjiang/env_ensforcing ? After you source g5_modules, it should be working. Then it avoids hassle to change the environment.

weiyuan-jiang (commented):
I have a question for this command:

 python  average_ensemble_forcing.py

1) it should be copied to somewhere during ldas_setup
2) there is hard coded location of the forcing.  The path should be passed in or changed manually

saraqzhang:
@weiyuan-jiang   just as you observed, we can use this opportunity to improve this enadas-coupling  option: (1) make the ens forcing path and Nmem as command line arg inputs to py (2) add cp the py code from util/ to exp/run/ in setup script. (3) test/use whatever the py version already in the ldas environment ( 2.7 ?) .

weiyuan-jiang:
OK. I will take care of this branch.

weiyuan-jiang:
@saraqzhang , I have updated this branch but I didn't test it. 
if LDAS_COUPLING == 2, you will need two extra parameters in your exec input file ( i.e. input file in for ldas_setup) :  
LADAS_FORCE_PATH_IN: 
LADAS_FORCE_PATH_OUT:

I guess LADAS_FORCE_PATH_OUT is related to MET tag, so I am wondering to connect the two

saraqzhang:
@weiyuan-jiang ,   it is related to MET tag, but not the whole story.  it would be  $FVHOME in  ensadas setup. maybe we can do 
ADASHOME:   ( $FVHOME user input) 
 LADAS_FORCE_PATH_IN = $ADASHOME/atmens/ensdiag 
LDAS_FORCE_PATH_OUT = $LADAS_FORCE_PATH_IN/ensavg,  and py creates this dir to put avg files in there. 
then later we make our input/met_forcing/forc pointed to  the location of LDAS_FORCE_PATH_OUT.

src/Applications/LDAS_App/ldas_setup
```diff
@@ -1438,30 +1446,42 @@ def _printExeInputKeys(rqdExeInpKeys):
...
+    print '#  (2) EXP_DOMAIN must be global CS grid as in ADAS exp    #'
     print '#                                                          #'
-    print '#  (3) MET_TAG  must be set to  ${ADAS_exp}__Nx+-          #'
-    print '#      MET_PATH must be set to  ${ADAS_exp}/lana/forc      #' 
-    print '#    After ldas exp setup, verify the following link:      #'        
-    print '#    ../input/met_forcing/forc -> ${ADAS_exp}/lana/forc    #'
+    print '#  (3) MET_TAG  must be set to [ADAS_EXPID]__Nx+-          #'
+    print '#      MET_PATH must be set as follows for                 #'
+    print '#        LADAS_COUPLING = 1:                               #'        
+    print '#          [ADAS_EXPDIR]/lana/forc                         #'
```
gmao-rreichle, line 1465: Are the "./lana/forc" and "./atmens/ensdiag/forc" sub-directories hardwired into the ADAS? 
The "./atmens/ensdiag/forc" sub-directory seems different from what used to be hard-coded in "average_ensemble_forcing.py".  Is the dir name actually correct here?

gmao-rreichle:
@saraqzhang : I made a few changes, primarily in the hope of clarifying the documentation.  I did not test the changes.  Please take a look a make sure I didn't introduce any errors.  Thanks.

src/Applications/LDAS_App/ldas_setup
saraqzhang: @gmao-rreichle   ./lana/ forc  is hardwired in ADAS central, it links to fvwork when the run is going.  ./atmens/ensdiag/forc  is not hardwired in ADAS, it gets created by the python code to hold the output of py averaging, so technically it does not have to be within ADAS dir.   An advantage of this specific path dir is  it is put within atmens existing stage dir structure, it stays there and gets archived along with atmens.

gmao-rreichle: @saraqzhang : Thanks for the info.  One more question going deeper into details... Is the ./lana/forc directory hardwired into _your_ version of the ADAS, or is it already in an official ADAS tag?  Perhaps this somehow made it into an official ADAS tag when Clara was working on the LADAS, in which case we should just work with it as is.  If the ./lana/forc directory is not already in an official ADAS tag, it may not survive the pre-merger code review, but we can deal with it then.  Either way, for now it's fine as is.

saraqzhang: @gmao-rreichle :  ./lana was in the  original ADAS tag ( in fvsetup,  make the dir).    ./lana/forc is created in the process of running g5adas  ( point to fvwork/ ), and removed when the run finishes ( fvwork is gone by then).

saraqzhang:
@gmao-rreichle :  I pulled the branch and tested setup , no problem. #347 
 I just made a minor change  (from 1 to 2 ) in ldas_setup comments:
print '#        LADAS_COUPLING = 2:

gmao-rreichle (approved):
@biljanaorescanin : Please test and merge when you get a chance.  Thanks!

tclune (approved):
CMake changes OK.

biljanaorescanin:
Branch passed Nightly testing as zero diff.

PR #349, opened 2021-01-08
use g5_modules's python

weiyuan-jiang:
The util's env should be consistent with g5_modules


weiyuan-jiang:
This is a very small changes.  Maybe it is not necessary. @saraqzhang . If you can incorporate the change to your PR, we can close this PR

weiyuan-jiang:
close. Will be incorporated into #347

PR #351, opened 2021-01-15
fix when there is no landpert restart file for simulations in cube-sphere tile space

weiyuan-jiang:
Add check if "LANDPERT_ INTERNAL_RESTART_FILE" exists

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90
@@ -983,7 +983,7 @@ contains
        endif
 
        call ESMF_CFIOStrTemplate(rst_fname, trim(adjustl(rst_fname_tmp)),'GRADS', xid = trim(id_string), stat=status)
-       inquire(file=rst_name, exist=f_exist)
+       inquire(file=rst_fname, exist=f_exist)
 
        if (index(rst_fname, 'NONE') == 0 .and. f_exist) then
 
--- a/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90
@@ -876,7 +876,7 @@ contains
     integer :: imjm(7), imjm_global(7) ! we need just the first 2
     integer :: model_dtstep
     integer :: land_nt_local,m,n, i1, in, j1, jn
-    logical :: IAmRoot
+    logical :: IAmRoot, f_exist
     integer :: ipert,n_lon,n_lat, n_lon_g, n_lat_g
     integer, allocatable :: pert_rseed(:)
     real :: dlon, dlat,locallat,locallon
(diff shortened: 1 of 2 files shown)


gmao-rreichle (commented):
Looks good to me on paper, but...

@weiyuan-jiang : Did you test this PR?  
If not, I think that @biljanaorescanin should test the branch first. 
However, the test is somewhat tricky because we need to edit the test in @mathomp4 's  directory by removing the links from @mathomp4 's test dir to the **landpert** restart files in @saraqzhang 's directory.  That is, to match the baseline output we need to ensure that the landpert GC is cold-started again.  Because this latext branch now always has the LANDPERT_INTERNAL_RESTART_FILE line in LDAS.rc, there must not be linked landpert restarts present, otherwise the cold-start would not kick in. 

BTW, we really should have copies of @saraqzhang 's input files in the test dir, not links.  If I remember correctly, only the files for the first four ens members are needed.  Maybe take this opportunity to copy the relevant restart files into the test dir (but not the landpert restarts...).

If everything checks out, @biljanaorescanin can then approve and merge.

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90
```diff
@@ -983,8 +983,9 @@ subroutine Initialize(gc, import, export, clock, rc)
        endif
 
        call ESMF_CFIOStrTemplate(rst_fname, trim(adjustl(rst_fname_tmp)),'GRADS', xid = trim(id_string), stat=status)
+       inquire(file=rst_fname, exist=f_exist)
```
gmao-rreichle, line 986: @weiyuan-jiang : One more thing... Should we restrict this "inquire()" to the Root node?  That is, replace this line with:
`if ( IAmRoot)  inquire(file=rst_fname, exist=f_exist)`
I suppose there's no need for all processors to send off their own inquiry to the file system.

gmao-rreichle: If we restrict `inquire()` to Root, then we'd also have initialize `f_exist = .true.` further up, I think.

weiyuan-jiang:
@gmao-rreichle , It is weird that I saw your comment in email but not here. I have made changes according to you suggestion.

biljanaorescanin:
@gmao-rreichle I've just tested the branch its zero diff. I will send email to @mathomp4 what needs to be removed from LDAS_Restarts for failing globalcs assim and gnuglobalcs assim tests to pass and cold start to kick in.

biljanaorescanin (approved):

gmao-rreichle:
Hi @biljanaorescanin : 
Thanks!  Just to clarify, did you test the branch before or after the latest commit by @weiyuan-jiang , which happened at 9:05am this morning, which was the result of a misunderstanding on my part and some back-and-forth.  If the test was before @weiyuan-jiang 's commit, we cannot say for sure that Matt's test will pass (after the adjustments to ensure a cold restart). Something to keep in mind for now.

biljanaorescanin:
@gmao-rreichle test was submitted at 16PM today. It was tested with the latest commit from @weiyuan-jiang.

Issue #359, opened 2021-01-28
Non-GEOS forcing data set to 0 in LDAS_move_new_force_to_old


gmao-jkolassa:
When using non-GEOS forcing data, the following forcing variables are set to 0 in LDAS_move_new_force_to_old.

Rainf_C
Rainf
Snowf
LWdown
SWdown
SWnet
PARdrct
PARdffs

These are passed with 0 values to LDAS_TInterpForcing, where some exit with 0 values as well. This causes unrealistic values in some model outputs.

The issue is likely a misplacement of the statement 'if (.not. GEOS_forcing) return' in  LDAS_move_new_force_to_old, which probably should be moved to before the above variables are set to zero. That is, when using non-GEOS forcing data, do nothing in  LDAS_move_new_force_to_old.

cc: @gmao-rreichle @biljanaorescanin @weiyuan-jiang @smahanam @rdkoster

gmao-jkolassa:
Posting the relevant code snippet for clarity:

```
     old_force%Rainf_C = 0.0
     old_force%Rainf   = 0.0
     old_force%Snowf   = 0.0
     old_force%LWdown  = 0.0
     old_force%SWdown  = 0.0
     old_force%SWnet   = 0.0
     old_force%PARdrct = 0.0
     old_force%PARdffs = 0.0
     
     if (.not. GEOS_forcing) return

     old_force%Rainf_C = new_force%Rainf_C
     old_force%Rainf   = new_force%Rainf
     old_force%Snowf   = new_force%Snowf
     old_force%LWdown  = new_force%LWdown
     old_force%SWdown  = new_force%SWdown
     old_force%SWnet   = new_force%SWnet
     old_force%PARdrct = new_force%PARdrct
     old_force%PARdffs = new_force%PARdffs
```

gmao-rreichle:
Here's how "move_met_force_obs_new_to_old" worked in LDASsa.

The following is based on the final LDASsa tag: reichle-LDASsa_m3-16_6_p5_SLES12.

In `clsm_ensdrv_main.F90`, the following block initializes the forcing variables:

```
        ! get initial forcing (note input date_time "force_time_old")
        ! check whether SWnet is available if needed (alb_from_SWnet)

        call get_forcing( force_time_old, force_dtstep,                        &
             met_path, met_tag, N_catf, tile_coord_f, met_hinterp,             &
             met_force_obs_tile_new_f, move_met_force_obs_new_to_old,          &
[...]

PR #361, opened 2021-01-29
bugfix LDAS_move_new_force_to_old(): removed incorrect set-to-0 of non-GEOS flux-type forcing

gmao-rreichle:
Addresses #359 

I am pretty sure that removing the lines "old_force%[xxx] = 0." at the beginning of the subroutine will fix the immediate issue with non-GEOS forcing data.

I have no idea why these lines were put there in the first place.  It is possible that removing them will somehow break something else.

@gmao-jkolassa : Please test the proposed solution in your use case with Fanwei's S2S forcing data.

@weiyuan-jiang : Please take a look at the source code to see if this makes sense to you.

@biljanaorescanin : Please run the usual set of GEOSldas tests.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -447,15 +447,6 @@ contains
      logical,intent(in) :: GEOS_forcing
      integer, intent(in) :: AEROSOL_DEPOSITION
 
-     old_force%Rainf_C = 0.0
-     old_force%Rainf   = 0.0
-     old_force%Snowf   = 0.0
-     old_force%LWdown  = 0.0
-     old_force%SWdown  = 0.0
-     old_force%SWnet   = 0.0
-     old_force%PARdrct = 0.0
-     old_force%PARdffs = 0.0
-     
      if (.not. GEOS_forcing) return
 
      old_force%Rainf_C = new_force%Rainf_C
@@ -477,6 +468,18 @@ contains
      new_force%PARdrct = nodata_generic
      new_force%PARdffs = nodata_generic
 
+     
+     ! [moved here from below, reichle, 28 Jan 2021]
+     ! treat Wind as flux when forcing with MERRA
+     if (MERRA_file_specs) then
+         old_force%Wind  = new_force%Wind
+         new_force%Wind  = nodata_generic
+     endif
+
+     ! not sure what exactly the following fields are and
+     ! whether it makes sense to include them here
+     ! - reichle, 28 Jan 2021
+     !
      if( AEROSOL_DEPOSITION /=0) then
 
         old_force%DUDP001 = new_force%DUDP001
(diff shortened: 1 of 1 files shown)


weiyuan-jiang:
I don't see the problem of of the coding. It is more like a science issue.  If it is non-GEOS forcing, would you replace the old one with new forcing?  How the old forcings eventually get updated?

gmao-rreichle:
> I don't see the problem of of the coding. It is more like a science issue. If it is non-GEOS forcing, would you replace the old one with new forcing? How the old forcings eventually get updated?

@weiyuan-jiang : It's not really a coding issue per se (ie, not a problem with initialization or allocation or such), but it's also not really a science issue.  It's more of a workflow and data processing issue.

For details, maybe take a look (again?) at my comment in the corresponding issue [#359](https://github.com/GEOS-ESM/GEOSldas/issues/359) about how this all worked in LDASsa.

Briefly, what we read in at a nominal time differs between "non-GEOS" and  "GEOS" datasets.
For example, say the simulation time is 0z and we want to advance the model to 1z.  To do this, we need the "new" air temperature at 1z and we need the fluxes (precip, radiation, ...) for 0-1z. 
When we read forcing data for "1z", we get the following from "get_forcing()": 
* GEOS: temperature at 1z and (backward-looking) fluxes for **0-1z**
* non-GEOS: temperature at 1z and (forward-looking) fluxes for **1-2z**

What we read is initially put into the variables that are associated with the "new" (or "next") time.
The convention in LDAS is that we always use fluxes stored in the variables associated with the "old" (or "previous") time.
No action is needed for non-GEOS data, which already contain the appropriate values in the "old" flux variable (because the driver was originally written for non-GEOS forcing).
But for GEOS we now need to simply copy the "new" values into the "old" variables to meet the LDAS convention.

The problem with the GEOSldas subroutine in question is that it **always** set the "old" flux values to zero.  This is an error in the translation from LDASsa to GEOSldas that might go back to Purnendu's first implementation of the MetForce GridComp.  We never verified that GEOSldas works with the (mostly outdated) non-GEOS forcing.  I'm not aware of anyone having used non-GEOS forcing data with GEOSldas until Jana pointed out her use case yesterday. 

The more I think about it, the more I'm confident that my branch is ok.  But it's a very tricky issue.  More eyes on the logic and more tests would be helpful, hence my request for support.  Also, it'd be good for more people to understand how this all works,  in case I get run over by a bus tomorrow.

cc: @biljanaorescanin @gmao-jkolassa @rdkoster

gmao-jkolassa:
I ran two experiments to test this fix using for my use case with the S2S forcing data.

The first test was a 1-day test run over a single tile to verify that the data being passed to LDAS_TInterpForcing correspond to the forcing data in the file read. That is, they are no longer set to 0 by LDAS_move_new_force_to_old (as expected).

I also ran a one month global experiment to verify that with this change implemented, the model output is now able to approximate the old LDASsa experiment using the same forcing data reasonably well.

biljanaorescanin:
As far as Nightly tests go this PR is zero diff.

gmao-rreichle:
Thanks @biljanaorescanin , @gmao-jkolassa and @weiyuan-jiang for taking a look at this.
@biljanaorescanin : Please go ahead and merge the PR.

biljanaorescanin (approved):

Issue #362, opened 2021-02-01
Run in DEBUG


biljanaorescanin:
If we compile develop with "./parallel_build.csh -debug" flag, during ldas_setup run we will get this error: 

>  Creating domain..., reading IncludeList and ExludeList if present...
 Done with domain_setup
 SURFLAY:    50.00000    
forrtl: error (63): output conversion error, unit -5, file Internal Formatted Write
Image              PC                Routine            Line        Source             
preprocess_ldas.x  0000000000F3B9DB  Unknown               Unknown  Unknown
preprocess_ldas.x  0000000000F8F9B0  Unknown               Unknown  Unknown
preprocess_ldas.x  0000000000F8D773  Unknown               Unknown  Unknown
preprocess_ldas.x  00000000004C6221  ldas_ensdrv_funct         132  LDAS_ensdrv_functions.F90
preprocess_ldas.x  000000000044D777  preprocess_ldas_r        1420  preprocess_ldas_routines.F90
preprocess_ldas.x  000000000041FC5D  preprocess_ldas_r         191  preprocess_ldas_routines.F90
preprocess_ldas.x  000000000041BC4B  MAIN__                     70  preprocess_ldas.F90
preprocess_ldas.x  000000000041B112  Unknown               Unknown  Unknown
libc-2.22.so       00002AAAB8414725  __libc_start_main     Unknown  Unknown
preprocess_ldas.x  000000000041B029  Unknown               Unknown  Unknown

PR #363, opened 2021-02-01
fix for debug flags

weiyuan-jiang:
A combination of some debug flags leads to an internal writing failure. This is a workaround to address #362

### DIFF
--- a/src/Components/GEOSldas_GridComp/Shared/LDAS_ensdrv_functions.F90
+++ b/src/Components/GEOSldas_GridComp/Shared/LDAS_ensdrv_functions.F90
@@ -76,7 +76,8 @@ contains
     character(*),         optional :: file_ext      ! default = '.bin'
 
     logical,              optional :: no_subdirs    ! default = .false.  
-    
+
+    character(len=*), parameter :: Iam = 'get_io_filename'    
     ! locals
     
     integer        :: tmp_option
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@weiyuan-jiang : 
I suggest a revised fix, see commit [443291b](https://github.com/GEOS-ESM/GEOSldas/commit/443291bf19872206ef316645de07a51a67210c42)
Please take a look.  (I did not yet make sure that the fix even compiles.)
I guess the ultimate source of the problem is that subroutine get_io_filename() was called with "option=3" and a "date_time" argument that did not have the pentad field set properly.  BUT, I don't see a call to get_io_filename() with "option=3" anywhere in develop.  Is this something that came up in a developer's new branch?

biljanaorescanin:
@gmao-rreichle I got this error today on not so up to date branch (October clone). But than same error was reproduced with current develop. @weiyuan-jiang  was able to reproduce it as well.

weiyuan-jiang (commented):
looks good to me

weiyuan-jiang:
The run didn't exit successfully. Can you post the run time error?

biljanaorescanin:
1:
sacct: error: slurm_persist_conn_open_without_init: failed to open persistent connection to rmdb5:6830: Connection timed out
sacct: error: slurmdbd: Sending PersistInit msg: Connection timed out
sacct: error: Problem talking to the database: Connection timed out

biljanaorescanin:
@gmao-rreichle @weiyuan-jiang  Just by re-running tests now GNUGLOBAL/assim passed. Its was obviously some glitch in post processing. Good news is this PR is zero diff now.

gmao-rreichle (approved):

PR #367, opened 2021-03-19
exit job script when GEOSldas.x run failed

weiyuan-jiang:
exit when run failed.

### DIFF
--- a/src/Applications/LDAS_App/lenkf.j.template
+++ b/src/Applications/LDAS_App/lenkf.j.template
@@ -362,7 +362,7 @@ while ( $counter <= ${NUM_SGMT} )
       set rc = -1
       echo GEOSldas Run Status: $rc
       echo "GEOSldas run failed, exit without post-processing"
-      exit
+      break
    endif
     
    
--- a/src/Applications/LDAS_App/lenkf.j.template
+++ b/src/Applications/LDAS_App/lenkf.j.template
@@ -361,7 +361,7 @@ while ( $counter <= ${NUM_SGMT} )
    else
       set rc = -1
       echo GEOSldas Run Status: $rc
-      echo "GEOSldas run failed, exit without post-processing"
+      echo "ERROR: GEOSldas run FAILED, exit without post-processing"
       exit
    endif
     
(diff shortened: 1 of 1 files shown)


src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -357,10 +357,13 @@ while ( $counter <= ${NUM_SGMT} )
    
    if( -e EGRESS.ldas ) then
       set rc = 0
+      echo GEOSldas Run Status: $rc
    else
       set rc = -1
+      echo GEOSldas Run Status: $rc
+      echo "GEOSldas run failed, exit without post-processing"
+      exit
```
gmao-rreichle, line 365: Do we really want to exit here?  If we do, there will not  be an attempt at post-processing.  I would remove the "exit" here and in the preceding line say something like: 
`echo "GEOSldas run FAILED, attempting post-processing..."`
Then also move the "if ... exit" statement to occur ~line 714 (see comment below).

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -732,6 +735,11 @@ EOF
 ## End of the while ( $counter <= ${NUM_SGMT} ) loop ##
 end
 
+if ($rc != 0) then
```
gmao-rreichle, line 738: If we drop the exit statement above, this if block should be moved to ~line 714, right?

src/Applications/LDAS_App/lenkf.j.template
weiyuan-jiang: This one should be "break" which comes out of the loop

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -357,10 +357,13 @@ while ( $counter <= ${NUM_SGMT} )
    
    if( -e EGRESS.ldas ) then
       set rc = 0
+      echo GEOSldas Run Status: $rc
    else
       set rc = -1
+      echo GEOSldas Run Status: $rc
+      echo "GEOSldas run failed, exit without post-processing"
+      break
```
gmao-rreichle, line 365: If we "break" here, don't we then immediately hit the "exit" statement in line 740? If this is the desired behavior, then we might as well "exit" right here, or not? 
My earlier thought was that we should try to finish out the current SGMT, and then have the if block with the rc-triggered exit in ~line 714, but on 2nd thought I'm not sure this makes much sense.
In summary, change "break" back to "exit" here and remove the new if block that was added ~Line 738?

gmao-rreichle (approved):

PR #371, opened 2021-04-06
Update GEOSldas to env 3.2.0, MAPL 2.6.4, cmake 3.3.7

mathomp4:
This PR brings the GEOSldas `develop` up to date with what GEOSgcm is or will be when @sdrabenh pulls in https://github.com/GEOS-ESM/GEOSgcm/pull/278.

This has three updates:
* ESMA_env v3.2.0
* MAPL 2.6.4
* ESMA_cmake v3.3.7

Of these, the latter one is essentially meaningless to the GEOSldas as it adds options to `parallel_build.csh` that only affects FV3. 

But, env and MAPL are more substantive. MAPL 2.6.4 I don't think should have any ill effects on LDAS. ESMA_env v3.2.0 is a big change in that it moves to ESMF 8.1.0. @biljanaorescanin did a test for me and found it zero-diff, but it's probably best to re-try the tests with this update to all three.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -5,13 +5,13 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  branch: test-esmf810-env
+  tag: v3.2.0
   develop: main
 
 cmake:
   local: ./@cmake
   remote: ../ESMA_cmake.git
-  tag: v3.3.6
+  tag: v3.3.7
   develop: develop
 
 ecbuild:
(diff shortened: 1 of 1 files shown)


gmao-rreichle (approved):
Ok with me but before merging perhaps @biljanaorescanin should test this again as a whole package per suggestion from @mathomp4

biljanaorescanin:
All tests pass, I can confirm its zero diff.

PR #372, opened 2021-04-07
recover stored_field for fft results

weiyuan-jiang:
Save the second part of FFT field to improve performance.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/land_pert.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/land_pert.F90
@@ -740,7 +740,14 @@ contains
     
     type(random_fields) :: rf
 
-    ! ---------------------------
+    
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
I'm running tests now and will make a science evaluation of the outputs.   See also #373

gmao-rreichle:
@weiyuan-jiang : your comment from issue #373 is correct:
>I have a feeling that it is wrong to save the second field for different variables because the initialization is different.
> call rf%initialize(rNlon, rNlat, 1., xCorr, yCorr, rdlon, rdlat )
> But it is ok for different ensemble members.

If we can't use a second random field for a given variable for another ensemble member at the same simulation time, perhaps we can remember the second field until the next time perturbations are generated?   I don't know much about the Fortran "save" statement, except that I remember someone saying it's not good practice...
Perhaps the parallel FFT will bring down the relative effort for the random field generator enough for it to no longer be an issue.

weiyuan-jiang:
closed. The perturbation field should not be stored this way.

Issue #373, opened 2021-04-07
implement parallel FFT and clean up perturbations subroutines


gmao-rreichle:
There are two tasks related to perturbations:

1. Implement parallel FFT to speed up computation of correlated random fields  It's not clear if this will be 0-diff.  Probably not.

2. Clean up (simplify) the perturbations code.  Currently, subroutines get_pert() and propagate_pert() exist in two flavors, "GEOSldas" and "LDASsa".   The key difference between the "GEOSldas" and "LDASsa" versions is that the latter has an additional array dimension for the ensemble members.  
If I remember correctly, the "GEOSldas" flavor is used for forcing and prognostics perturbations, and the "LDASsa" version is used for perturbing observations.  
Would it be possible to merge the subroutines again into just one version that always includes the ensemble member dimension, and then call this merged subroutine for just one member when the "GEOSldas" flavor is needed?  This may require some reshuffling of the array dimensions in the calling subroutines.  
Having just one subroutine for get_pert() and one for propagate_pert() would greatly simplify maintenance of the code.  
This change should be 0-diff.  

Perhaps it would be best to start with item 2 and then implement the FFT?

weiyuan-jiang:
I have a feeling that it is wrong to save the second field for different variables because the initialization is different. 

call rf%initialize(rNlon, rNlat, 1., xCorr, yCorr, rdlon, rdlat )

But it is ok for different ensemble members.

gmao-rreichle:
addressed with #418 and #376

PR #376, opened 2021-04-09
cleanup pert subroutines

weiyuan-jiang:
This PR addresses part 2 of issue #373 to clean up pert subroutines

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
@@ -3339,8 +3339,8 @@ contains
     !
     ! get gridded perturbations
     
-    allocate(Obs_pert_ntrmdt(N_obs_param,pert_grid_lH%N_lon,pert_grid_lH%N_lat,N_ens))
-    allocate(Obs_pert_grid(  N_obs_param,pert_grid_lH%N_lon,pert_grid_lH%N_lat,N_ens))
+    allocate(Obs_pert_ntrmdt(pert_grid_lH%N_lon, pert_grid_lH%N_lat, N_obs_param, N_ens))
+    allocate(Obs_pert_grid(  pert_grid_lH%N_lon, pert_grid_lH%N_lat, N_obs_param, N_ens))
 
     call get_pert(                                                        &
          N_assim_species, N_ens,                                          &
@@ -3348,8 +3348,8 @@ contains
          dtstep,                                                          &
          obs_pert_param,                                                  &
          Pert_rseed,                                                      &
-         Obs_pert_ntrmdt(1:N_assim_species,:,:,:),                        &
-         Obs_pert_grid(  1:N_assim_species,:,:,:)              )
+         Obs_pert_ntrmdt(:,:,1:N_assim_species,:),                        &
+         Obs_pert_grid(  :,:,1:N_assim_species,:)              )
 
     ! clean up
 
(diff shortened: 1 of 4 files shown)


src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -3339,17 +3339,17 @@ subroutine get_obs_pert( N_ens, N_obs, N_obs_param,                  &
     !
     ! get gridded perturbations
     
-    allocate(Obs_pert_ntrmdt(N_obs_param,pert_grid_lH%N_lon,pert_grid_lH%N_lat,N_ens))
-    allocate(Obs_pert_grid(  N_obs_param,pert_grid_lH%N_lon,pert_grid_lH%N_lat,N_ens))
+    allocate(Obs_pert_ntrmdt(pert_grid_lH%N_lon, pert_grid_lH%N_lat, N_obs_param, N_ens))
```
gmao-rreichle, line 3342: Here  and in the next line, do we need to allocate with "N_obs_param" when in the call to get_pert() below we're only using "N_assim_species"?
By construction, N_assim_species <= N_obs_param. 
I'm pretty sure we can save a bit of memory.  Can you please double-check?

gmao-rreichle:
I added a few comments and cleaned up a bit.  I am testing now.  Will update with results later.

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/land_pert.F90
```diff
@@ -503,7 +257,7 @@ subroutine LDASsa_get_pert(                         &
 
           do i=1,pert_grid_l%N_lon
 
-             call adjust_mean(pert_grid_l%N_lat, N_ens, Pert_ntrmdt(m,i,:,:) )
+             call adjust_mean(pert_grid_l%N_lat, N_ens, Pert_ntrmdt(i,:,m,:) )
```
gmao-rreichle, line 260: A silly question... We are passing a 4-dim array with two singleton dimensions into adjust_mean(), which expects a 2-dim array.  Does Fortran automatically "squeeze" the singleton dimensions and effectively present the desired 2-dim array to adjust_mean? 
In the new branch, the non-singleton dimension are no longer consecutive, as they are in the original code.

gmao-rreichle:
@weiyuan-jiang :
I successfully tested "conus", "global" and "globalcs" with Intel.   Didn't test Gnu.
One more question: If I understand the modifications correctly, the change in the order of array dimensions shouldn't impact the pert restart files, correct?  I think the Nightly assim test uses pert restarts, so everything should be ok and tested, but I thought I should mention pert restarts just in case I misunderstood or missed something.
Besides that, perhaps we still want to change the allocation of Obs_pert_ntrmdt and Obs_pert_grid. See inline comment above.
Please review what I've changed and let me know, thanks!

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/land_pert.F90
weiyuan-jiang: It may be faster if if the two-D is the first two because of the continuous memory.  Fortran won't squeeze it but would remember the stride between elements.

weiyuan-jiang:
It won't affect the landpert restart order because all I changed is the 'LDASsa' style order that has nothing to do with restart. I have changed the allocation sizes of  Obs_pert_ntrmdt and Obs_pert_grid.

weiyuan-jiang:
The purpose to change the "LDASsa" style order is to preserve the "GEOSldas"  style order that would affect the restarts

gmao-rreichle:
@biljanaorescanin :
When you get a chance, please run a full test of this branch (incl GNU) and, if successful, go ahead with the merge, unless there are objections from @weiyuan-jiang.  If anything is unclear, we can discuss on Wednesday.  Thanks.

weiyuan-jiang:
I have tested it. So it should be good to merge. @biljanaorescanin @gmao-rreichle

biljanaorescanin:
All tests passed. ( it took a while discover is so slow)

biljanaorescanin (approved):

PR #377, opened 2021-04-10
Adding ERA5 as Met Forcing option

biljanaorescanin:
Reader for ERA5 files shared on DISCOVER by LIS group.


biljanaorescanin:
All tests passed.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -1898,7 +1928,295 @@ subroutine get_conus_netcdf(date_time, met_path, N_catd, tile_coord, &
...
+       this_lon = tile_coord(k)%com_lon
+       this_lat = tile_coord(k)%com_lat
+       
+       ! add small offsets to avoid unpredictable assignment of
+       ! regularly spaced tiles (such as from the EASEv2 tile space)
+       ! to ERA5 grid cells along certain lat/lon values
+       ! (that is, make it easier to reproduce the mapping done here
+       !  in separate postprocessing analysis scripts)
+       
+       this_lon = this_lon + 0.0001
```
gmao-rreichle, line 2064: An if statement should be added here that restricts the offsets to simulations when the tiles are essentially identical to regular lat/lon grid cells, such as in the EASE grid tile spaces.  We do not have ready access to the kind of tile space we're dealing with here, so the if condition is not trivial.  This requires a bit more digging.  Weiyuan can probably help.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -1898,7 +1928,295 @@ subroutine get_conus_netcdf(date_time, met_path, N_catd, tile_coord, &
...
+       this_lat = this_lat + 0.0001
+       
+       i_ind(k) = ceiling( (this_lon - era5_grid_ll_lon)/era5_grid_dlon )
+       j_ind(k) = ceiling( (this_lat - era5_grid_ll_lat)/era5_grid_dlat )
+
+       ! NOTE: For a "date line on center" grid and (180-dlon/2) < lon < 180
```
gmao-rreichle, line 2070: The ERA5_LIS grid is **not** a date-line-on-center grid.  Is this actually needed here?  I wouldn't think so.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -1898,7 +1928,295 @@ subroutine get_conus_netcdf(date_time, met_path, N_catd, tile_coord, &
...
+
+    if(root_logit) write (logunit,*) 'get ERA5 LIS file forcing data lat/lons from ' // trim(fname)
+
+    ierr = NF90_OPEN(fname,NF90_NOWRITE,ncid)
+    if (ierr/=0) then
+       err_msg = 'error opening netcdf file'
+       call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)
+    end if
+
+    ! read Longitude and Latitude from nc file
```
gmao-rreichle, line 2127: I don't understand why we need the lat/lon values as a gridded variable.  We already know the i,j coordinates for each tile on the 2D ERA5_LIS grid (ie, i_ind, j_ind), and we shouldn't need the lat/lon values to reassemble the forcing data from their compressed 1d array onto the 2d ERA5_LIS grid.  I think the entire section dealing with lat/lon values can be deleted.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -1898,7 +1928,295 @@ subroutine get_conus_netcdf(date_time, met_path, N_catd, tile_coord, &
...
+          tmp_grid(land_i_era5(n),land_j_era5(n)) = tmp_vec(n)
+       end do
+
+       ! interpolate to tile space
+       do k=1,N_catd
+          ! if the GEOS tile is not within the 0.125*sqrt(2) deg of the matching ERA5
+          ! tile, asign a missing value to it (this will trigger Exclude list flagging)
+          ! otherwise, feed with colocated tmp_grid element
+
+          geos2era_tile_SQdistance =                                             &
```
gmao-rreichle, line 2177: I don't understand why we need the SQdistance threshold.
If we initialize tmp_grid to nodata (which was missing!), then if a tile ends up in a grid cell that is outside of the ERA5_LIS compressed domain, its forcing will automatically be assigned a nodata values.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
weiyuan-jiang: quote : "that is, make it easier to reproduce the mapping done here in separate post processing analysis scripts)". I think the post processing tools should change, not here.

gmao-rreichle: @weiyuan-jiang , the offset needs to be applied both here and in the post-proc tools.  Let me restate the problem: Assume you have data on a grid at lat=0 and lat=1 and the tile is at exactly lat=0.5.  Then the nearest neighbor could be either lat=0 or lat=1, depending on roundoff.  The problem is that we cannot construct post-proc tools that match exactly the roundoff in GEOSldas.  The offset avoids the problem from occurring by intentionally and predictably going with the nearest neighbor in one direction, which can then be implemented in the same way in the post-proc tools.  I hope this makes sense now.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
weiyuan-jiang: Now I see the problem.  Use function NINT  instead of ceiling can solve the problem

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -1898,7 +1928,295 @@ subroutine get_conus_netcdf(date_time, met_path, N_catd, tile_coord, &
...
+             (tile_coord(k)%com_lon+0.0001 - era5_lon2D(i_ind(k),j_ind(k)))**2 + &
+             (tile_coord(k)%com_lat+0.0001 - era5_lat2D(i_ind(k),j_ind(k)))**2
```
weiyuan-jiang, line 2179: This ( +0.0001)  is quite awkward  to me.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
weiyuan-jiang: quote : " intentionally and predictably going with the nearest neighbor in one direction" . Ok, then nint may not be the solution.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: > The ERA5_LIS grid is **not** a date-line-on-center grid. Is this actually needed here? I wouldn't think so.
Probably not needed, but I need to take time to do a deeper investigation. Initially, I concluded this should not cause a problem if stays and never came back to revisit (will do).

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: True. I'll see to remove this. The reason why I needed it was probably the missing initialization in the first place, but when I was looking into the continent coasts problem this is how I identified the problem. Now when cleared, I think we can get rid of it. I'll test it and update it.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: I agree, but it was an easy and quick way to fix the issue. I am open to suggestions.

gmao-rreichle: This block is almost certainly going away, so should not be an issue. 
Further up, the +0.0001 is going to be needed unless someone has a better solution.
But I still think we should at an "if (EASE grid tile space)" condition if that's possible.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: I took a closer look, ran a couple of tests, confirmed the i_ind never exceeds the "era5_grid_N_lon" value. This part can be removed.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: Tested. Confirmed. Will be removed in branch update.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: This entire section will be removed, since the distance match is not needed anymore.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
biljanaorescanin: I removed this section.

gmao-rreichle:
@biljanaorescanin @weiyuan-jiang 
I pushed two commits onto @biljanaorescanin 's branch that should address issue #378. 
I successfully tested conus and global (both model and assim) with Intel.  Not the complete set of tests, but good enough for now. 
I want to do more cleanup and eliminate SWnet from met_force_type.  Not sure when I can do this, but once I find the time it should be straightforward.
In the meantime:
@weiyuan-jiang , please take a look at my two commits 
[de44185](https://github.com/GEOS-ESM/GEOSldas/pull/377/commits/de4418539554836ff4849181172c1665282e4d14)
[67cc80f](https://github.com/GEOS-ESM/GEOSldas/pull/377/commits/67cc80fde746a0e85eca225218df9c1c213b4289)
and let me know if this makes sense to you.  The commits include what you had on your separate [branch](https://github.com/GEOS-ESM/GEOSldas/tree/feature/wjiang%2Fclean_up_forcing_par) in response to #378.  When you look at the second commit, make sure to ignore white space.
@biljanaorescanin , please continue debugging the ERA5_LIS reader.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -317,84 +331,88 @@ subroutine get_forcing( date_time, force_dtstep, met_path, met_tag,        &
...
+    ! where PAR is not available for any of the above reasons:
+    ! - assume half of SWdown is photosynthetically active
+    ! - assume half of PAR is direct, half diffuse
     
-    ! check for nodata values, possibly fill with neighboring data
+    tol = abs(nodata_forcing*nodata_tolfrac_generic)
+    
+    where (                                                                       &
+         (abs(met_force_obs_tile_new%PARdrct-nodata_forcing) < tol)  .and.        &
+         (abs(met_force_obs_tile_new%SWdown -nodata_forcing) < tol)            )
```
weiyuan-jiang, line 371: I think this is incorrect.   If both are true, then there is no data

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -287,130 +270,60 @@ subroutine get_forcing( date_time, force_dtstep, met_path, met_tag,        &
...
        ! (default is to limit Qair so that it does not exceed Qair_sat)
        ! reichle+qliu,  8 Oct 2008
        ! 
        ! likewise for "unlimited_LWdown=.true."
        ! reichle, 11 Feb 2009       
        
-       call repair_forcing( N_catd, met_force_obs_tile_new, &
-            echo=.true., tile_coord=tile_coord,             &
-            unlimited_Qair=.true., unlimited_LWdown=.true. )
+       unlimited_Qair                 = .true.
```
weiyuan-jiang, line 332: move the two lines after get_GEOS so it is consistent with ge_ERA5_LIS and shows they are not used in the calls

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: Good catch, thanks!  SWdown must NOT be no-data.  I'll fix it.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: Thanks! I rearranged the statements (and a couple more things) for clarity.

gmao-rreichle:
@biljanaorescanin @weiyuan-jiang 

FYI, I merged the SWnet cleanup into the branch associated with the present pull request (`feature/borescan_era5`). For details, see #386.   Tests were 0-diff (except for missing SWnet in "lfs" HISTORY output).  When the present PR is merged into "develop", we need to re-baseline the "model" cases in Matt's nightly tests (and probably also adjust the input HISTORY.rc files).

The nightly tests only exercise the GEOS forcing reader.  I hope we can soon sort out the "vertical stripes" in the case with the ERA5 reader.

biljanaorescanin:
Vertical stripes resolved. Those were caused by a non-matching number of tiles in the run and the ancillary albedo file, not by the ERA5 reader.

gmao-rreichle:
Hi @biljanaorescanin, 
I added a couple more commits, mostly cleanup of obsolete variable declarations, comments, empty lines, indentation, etc:
[8ef7f1e](https://github.com/GEOS-ESM/GEOSldas/pull/377/commits/8ef7f1e7dd88140653f178f87798c6e94fb19010)
[2808d26](https://github.com/GEOS-ESM/GEOSldas/pull/377/commits/2808d26d176930437db4b11af3c4153d7dedc2c2)
The first commit also includes a simplification of the integer calculations to go from the 1-dim `p2g` index to 2-dim `i,j` indices.  I checked this calculation against yours with a small, separate test program.  Nevertheless, it's important that you run a brief test with ERA5 forcing to make sure I didn't mess up the reader.  (Sorry in advance if I did!)  
Also, just to be safe, perhaps run an abbreviated version of the "nightly" tests.  Because of the nature of the changes, this PR can almost be considered trivially 0-diff.  
Let me know once you have the test results, then I can approve the PR.
cc: @weiyuan-jiang

biljanaorescanin:
@gmao-rreichle Thanks for the cleanup it does look neater. 
I've tested the changes in few simple runs and they are the same as before. Nightly test also passed.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: Let's leave this for another day...

gmao-rreichle (approved):

PR #383, opened 2021-04-22
parallel fft

weiyuan-jiang:


weiyuan-jiang:
https://github.com/GEOS-ESM/GEOSldas/blob/49ae464151f57df4d258a16df13ea87606444ce6/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/GEOS_LandPertGridComp.F90#L1069-L1072

If the pgrid_l is passed in above call, some processors without tiles will get npert = 0.  In the parallel fft, the collective call "AllToAll" would fail.   Can pfrid_f be passed in here?

weiyuan-jiang:
This PR is replaced with #418 . There is no performance improvement if all processors do parallel FFT together

PR #392, opened 2021-05-05
    use  assim alarm for centered_update option

saraqzhang:
1. Add centered_update option to the setting of the assim alarm  in GEOSlandassimGC Initialize. 
2. Remove  old code structure for centered_update option inhered from LDASsa.

 
- When centered_update = .true. in LDASsa_SPECIAL_inputs_ensupd.nml ,   analysis is carried out at the time centered in 3h forecast window ( for instance at 1:30Z  in 0 to 3Z model integration window) . 
- All analysis outputs are written at the corresponding time. For files controlled by MAPL,  in History.rc the ref.time need to be adjusted accordingly ( 01:30  instead of 00:00) .

### DIFF
--- a/src/Applications/LDAS_App/GEOSldas_HIST.rc
+++ b/src/Applications/LDAS_App/GEOSldas_HIST.rc
@@ -340,6 +340,9 @@ COLLECTIONS:
                                'TB_LAND_1410MHZ_40DEG_VPOL' , 'LANDASSIM'   , 'tb_v'                              ,
 			       ::
 
+# For catch_progn_incr, *.frequency and *.ref_time must be consistent with the LDAS.rc resource
+# parameters LANDASSIM_DT and LANDASSIM_T0.
+
   catch_progn_incr.descr:       'Tile-space,3-Hourly,Instantaneous,Single-Level,Assimilation,Ensemble-Average Land Prognostics Increments',
   catch_progn_incr.template:    '%y4%m2%d2_%h2%n2z.bin',
   catch_progn_incr.mode:        'instantaneous',
@@ -372,6 +375,9 @@ COLLECTIONS:
                                'SNDZN3_INCR'    , 'LANDASSIM'      ,
                          ::
 
+# For lndfcstana, *.frequency and *.ref_time must be consistent with the LDAS.rc resource
+# parameters LANDASSIM_DT and LANDASSIM_T0.
+
   inst3_1d_lndfcstana_Nt.descr:       'Tile-space,3-Hourly,Instantaneous,Single-Level,Assimilation,Ensemble-Average Land Forecast and Analysis Diagnostics',
   inst3_1d_lndfcstana_Nt.template:    '%y4%m2%d2_%h2%n2z.bin',
   inst3_1d_lndfcstana_Nt.mode:        'instantaneous',
@@ -388,6 +394,9 @@ COLLECTIONS:
                                      'TPSURF_ANA'  , 'LANDASSIM' , 'TSURF_ANA'       ,
                                      'TSOIL1_ANA'  , 'LANDASSIM' , 'TSOIL1_ANA'      ,
                         ::
+
+# For lndfcstana, *.frequency and *.ref_time must be consistent with the LDAS.rc resource
+# parameters LANDASSIM_DT and LANDASSIM_T0.
 
   inst3_2d_lndfcstana_Nx.descr:       '2d,3-Hourly,Instantaneous,Single-Level,Assimilation,Ensemble-Average Land Forecast and Analysis Diagnostics',
   inst3_2d_lndfcstana_Nx.template:    '%y4%m2%d2_%h2%n2z.nc4',
--- a/src/Applications/LDAS_App/GEOSldas_LDAS.rc
+++ b/src/Applications/LDAS_App/GEOSldas_LDAS.rc
@@ -71,6 +71,35 @@ MET_HINTERP: 1
 #
 LAND_ASSIM: NO
 
+# ---- Specify land assimilation times (when "LAND_ASSIM: YES")
+#
+#  LANDASSIM_DT : land analysis time step        (seconds) 
+#  LANDASSIM_T0 : land analysis "reference" time (hhmmss)
+# 
+#  LANDASSIM_T0 ("T0") and LANDASSIM_DT ("DT") define an infinite sequence of land analysis times:
+#
+#     ..., T0-3*DT, T0-2*DT, T0-DT, T0, T0+DT, T0+2*DT, T0+3*DT, ...
+#
+#  There is never a land analysis at the restart time.  Otherwise, the land analysis times are
+#  independent of the restart time.  There may be a land analysis at the final time.
+#
+#  LANDASSIM_DT must be <=86400s, be a multiple of HEARTBEAT_DT, and evenly divide a day.
+#  Consequently, only HHMMSS information is needed for LANDASSIM_T0.
+#
+#  Examples:
+#   T0=013000, DT=10800s --> land analysis whenever model time reaches 1:30z, 4:30z, ..., 22:30z.
+#   T0=163000, DT=10800s --> land analysis whenever model time reaches 1:30z, 4:30z, ..., 22:30z.
+#   T0=120000, DT=21600s --> land analysis whenever model time reaches 0z, 6z, 12z, and 18z.
+#
+#  LANDASSIM_DT and LANDASSIM_T0 work almost but not quite like "frequency" and "reftime" from MAPL
+#  HISTORY.  The difference is that MAPL HISTORY will not write output until "reftime" even when
+#  (reftime - restarttime)/frequency > 1.  E.g., if frequency=3h, reftime=9z, and restarttime=0z,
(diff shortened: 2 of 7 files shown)


gmao-rreichle:
Thanks, @saraqzhang !     @weiyuan-jiang, please take a look when you get a chance.

The changes look ok to me.  I created a new branch with some additional minor cleanup (feature/saraqzhang/centerupdate_RR).  I didn't want to commit my cleanup to Sara's branch in case I misunderstood something.   You can see my changes in the following difference view:
https://github.com/GEOS-ESM/GEOSldas/compare/feature/saraqzhang/centerupdate..feature/saraqzhang/centerupdate_RR
@saraqzhang : If you're ok with the cleanup, you can just merge my branch into yours.

Assuming the cleanup is fine, I think we still need to address the following things.  @weiyuan-jiang should be involved in this:

1) I may be mistaken, but it looks like "LandAssimDTstep" isn't read from the ensupd nml or LDAS resource file.  Rather, it's set to a default of 10800 seconds, and "dtstep_assim" from the ensupd nml file is ignored.  This needs to be fixed.  We need to have the assimilation time step explicitly in the resource file (and remove it from the ensupd nml file), or we need to get it from the ensupd nml file.

2) Not sure if/how to address the HISTORY.rc changes required for centered_update=.true.
In `GEOSldas_HIST.rc`, I think the "frequency" and "ref_time" of the `catch_progn_incr` and `lndfcstana` Collections depend on "LandAssim_DTstep" and "centered_update".  Not sure if  we can or want to try and modify this in pre-processing.  Note that the file name convention for these Collections **also** depends on "LandAssim_DTstep" (e.g., the "3" in "inst3_2d_lndfcstana_Nx" is for 3 hours, i.e., LandAssim_DTstep=10800s).  This may need to be left to the user for now.  But I'm open to ideas. 

3) Perhaps most importantly, the current design of the branch is a continuation of the LDASsa conventions.  The assumption is that a land analysis happens either at the end of "LandAssim_DTstep" or, with centered_update=.true., half-way through the "LandAssim_DTstep", and then again every "LandAssim_DTstep".  Perhaps a better design would be to replace the "centered_update" nml variable with a "LandAssim_RefTime" resource variable, which would mimic the MAPL HISTORY conventions for time steps and reference times.

weiyuan-jiang:
1) When Landassim_dtstep is used,  dtstep_assim was not used any more. I forgort to clean that up.  Oviously it is confusing and Sara picked up dtstep_assim. We need to clean that up and just pick one.
2) Can we template the history collections' name ? It doesn't seem easy at this point.
3) I think it is a good idea to add landassim_reftime

saraqzhang:
@weiyuan-jiang  @gmao-rreichle 
per our discussion I will make further changes : 
(1)  replace "centered_update" by LandAssim_refTime , and remove it from nml , add it to LDAS.rc. 
(2)  in GC Initialize , use MAPL_GetResource to get the value of LandAssim_refTime  (with default  ) 
(3)  create assim. alarm with information of LandAssim_DTstep  and LandAssim_refTime. 

Questions about  LandAssim_DTstep and dtstep_assim: 
(1) I could not find 'LandAssim_DTstep'  in  LDAS.rc.  In GC code  MAPL_GetResource must have been using the default.  Did I miss something here? 
(2) if we remove dtstep_assim from nml,  should  the setup prompt to put LandAssim_DTstep in LDAS.rc  in case of a different value from the default? 
(3) dtstep_assim is passed into various subroutines via argument. To keep the changes at a minimum,  we should keep the variable in the code and make dtstep_assim =  LandAssim_DTstep  if dtstep_assim is removed from nml?

gmao-rreichle:
@saraqzhang : see inline responses

> per our discussion I will make further changes :
> (1) replace "centered_update" by LandAssim_refTime , and remove it from nml , add it to LDAS.rc.
> (2) in GC Initialize , use MAPL_GetResource to get the value of LandAssim_refTime (with default )
> (3) create assim. alarm with information of LandAssim_DTstep and LandAssim_refTime.

Re. (1), I don't think we would add "centered_update" to LDAS.rc.  Rather, based on (3), we should have "DTSTEP" and "REFTIME" in LDAS.rc, and delete "dtstep_assim" and "centered_update" from the nml file.  I don't think the switch "centered_update" is needed anymore when we have "DTSTEP" and "REFTIME".   Note that "centered_update" is passed into "get_enkf_increments()" but is not used there, so can be removed from the interface. 

(2) and (3) sound good.

> (1) I could not find 'LandAssim_DTstep' in LDAS.rc. In GC code MAPL_GetResource must have been using the default. Did I miss something here?

There is no entry for "DTSTEP" in LDAS.rc yet, but we need to add it.

> (2) if we remove dtstep_assim from nml, should the setup prompt to put LandAssim_DTstep in LDAS.rc in case of a different value from the default?

There is already a call Get_MAPL_Resource() in the LandAssim GC that queries the rc file for DTSTEP.  This should be sufficient. (In addition, we need such a call for "REFTIME")

> (3) dtstep_assim is passed into various subroutines via argument. To keep the changes at a minimum, we should keep the variable in the code and make dtstep_assim = LandAssim_DTstep if dtstep_assim is removed from nml?

We should remove "dtstep_assim" from the nml file and subroutine read_ens_upd_inputs().  Simply pass "DTSTEP" (obtained via Get_MAPL_Resource()) into get_enkf_increments().   No need to rename the variable "dtstep_assim" within the other F90 files.

Note that the sanity checks for the consistency of dtstep_assim and model_dtstep need to be moved out of read_ens_upd_inputs() and into the LandAssim GC F90 file, probably best right after we read DTSTEP via Get_MAPL_Resource().

Note also that we currently set LandAssim_DTstep and dtstep_assim independently to 10800, and both are used (contrary to my first impression that dtstep_assim was ignored). It all works as long as the two dt values are consistent, but of course the cleanup described above is still necessary.

saraqzhang:
@gmao-rreichle 
 yes that is what I meant:   get rid of  centered_update, install LandAssim_refTime 
 will remove dtstep_assim  at GC level and in nml , pass "DTSTEP" to subroutines .

weiyuan-jiang:
@gmao-rreichle Do we need ref_date?  It includes year, month and date. ref_time includes hour, minute and second.

gmao-rreichle:
@weiyuan-jiang : we only need reftime.   I think in practice landassim_dtstep must be <= 24 hours.

weiyuan-jiang:
I am wondering if there is such situation that GEOSldas starts data assimilation a couple days later

gmao-rreichle:
@weiyuan-jiang : we never used such a scenario.  it should be so rare that users can just restart with a different config at the cutover date.  let's just work with ref_time for now.  but i appreciate the comments.  always good to think out of the box.

gmao-rreichle:
I merged the changes from PR [#397 ](https://github.com/GEOS-ESM/GEOSldas/pull/397) via PR [#404](https://github.com/GEOS-ESM/GEOSldas/pull/404):
* replaced "dtstep_assim" and "centered_update" with MAPL resource parameters "LANDASSIM_DT" and "LANDASSIM_T0"
* unrelated minor fix: made declaration of DataTmp consistent with DataPrv, DataNxt in GEOS_MetforceGridComp.F90 (potential GCC issue)
@biljanaorescanin : Please run a full test of this PR?
@saraqzhang : Please re-test the proper functioning of LANDASSIM_DT and LANDASSIM_T0 to replicate the "center_update" configuration needed for the LADAS.
cc: @weiyuan-jiang

saraqzhang:
@gmao-rreichle @weiyuan-jiang 
tests successful  (1) the proper functioning of LANDASSIM_DT and LANDASSIM_T0 to replicate the "center_update" configuration ( LDAS.rc LANDASSIM_T0: 013000)   (2)  with default setting of LANDASSIM_DT and LANDASSIM_T0

biljanaorescanin:
Current summary of the tests: Intel assim tests fail and all GNU test fail for this PR. 
Fail in Intel assim runs - problem is the SPECIAL ensupd nml file.  The test still contains "dtstep_assim" in the file, but this variable has been removed from the namelist. 
We are still looking into why GNU model build is failing.

gmao-rreichle:
@biljanaorescanin 
I just pushed a commit that will hopefully fix the GNU error.  I also added a couple of comments based on @saraqzhang's feedback. 
Please try again for GNU.  No need to retest Intel if it works for GNU.  I expect the GNU assim tests to fail for the same reason as the Intel test (ie, outdated SPECIAL nml file).  But Sara successfully verified 0-diff in her environment (in addition to successfully verifying that the new functionality works for her purpose).  
If the GNU build and model tests succeed, we can take a leap of faith:

1. merge the branch, and 
2. ask @mathomp4  to remove the "dtstep_assim" line from the SPECIAL nml files in his test directory.

If we can do both this afternoon and all goes well, the tests will all PASS tonight.  If they don't, I'll take the blame. 
cc: @weiyuan-jiang

biljanaorescanin:
Update on testing summary; we fail only assim run tests, same error mentioned before, outdated SPECIAL nml file. 
@mathomp4 this is ready for change @gmao-rreichle suggested on previous comment.

gmao-rreichle (approved):

PR #391, opened 2021-05-05
Feature/jpark50/#390 AODAS Forcing Reader

jpark50:
Adding the AODAS reader in the LDAS_Forcing.F90

PR #395, opened 2021-05-05
Feature/jpark50/#394 AODAS FOrcing Reader

jpark50:
Adding the AODAS Forcing reader in LDAS_Forcing.F90

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -289,13 +289,7 @@ contains
 
          call get_BC_AODAS( date_time_tmp, met_path, N_catd, tile_coord, &
               met_hinterp, met_force_obs_tile_new, nodata_forcing)
-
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@jpark50 

Thanks for creating a new pull request.  However, I hadn't asked for a new pull request.  The old PR [#391](https://github.com/GEOS-ESM/GEOSldas/pull/391)  would have worked just fine after redirecting the merge to the "develop" branch as described in my comments there, which you did.  I'm a bit confused why there's a new pull request.

Also, and much more importantly, your branch as it stands now cannot possibly work.  In the current develop branch, the %SWnet field has been removed from the "met_force" structure.  I pointed this out in emails and in our Teams chat:
image
It looks like you never 0-diff tested or even compiled this branch as required in Step 3 of the instructions for creating GEOSldas pull requests:
https://github.com/GEOS-ESM/GEOSldas/wiki/Protocol-for-GEOSldas-pull-requests
Note also that the calls to check_nodata_forcing() and repair_forcing() are now done after the if block that selects the individual forcing readers. 

cc: @lcandre2

gmao-jkolassa:
This PR is now addressed as part of PR #400 .

PR #397, opened 2021-05-07
remove dtstep_assim and centered_update from nml

weiyuan-jiang:
1)remove dtstep_assim and centered_update from nml
2)add LANDASSIM_REFTIME
3) by default, LANDASSIM_REFTIME is CurrentTime or Start time

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -1046,7 +1046,7 @@ contains
     type(ESMF_TimeInterval)      :: LandAssim_DT, one_day
     integer                      :: LandAssim_T0, LandAssimDTstep
     type(ESMF_TimeInterval)      :: ModelTimeStep 
-    type(ESMF_Time)              :: rrTime, pertSeedTime
+    type(ESMF_Time)              :: pertSeedTime
 
     ! locals
     type(MAPL_MetaComp), pointer :: MAPL=>null()
@@ -1209,7 +1209,20 @@ contains
     ! compute *earliest* AssimTime that is *greater* than CurrentTime:
     
     AssimTime = AssimTime + (INT((CurrentTime - AssimTime)/LandAssim_DT)+1)*LandAssim_DT
+    
+    ! create LandAssimAlarm
 
+    LandAssimAlarm = ESMF_AlarmCreate(                                          &
+         clock,                                                                 &
+         name='LandAssim',                                                      &
+         ringTime=AssimTime,                                                    &
+         ringInterval=LandAssim_DT,                                             &
+         ringTimeStepCount=1,                                                   &
+         sticky=.false.,                                                        &
+         rc=status                                                              &
+         )
+    _VERIFY(status)
+    
     ! ------------------------------------
     
     call ESMF_UserCompGetInternalState(gc, 'TILE_COORD', tcwrap, status)
@@ -1343,26 +1356,6 @@ contains
     call MPI_BCAST(out_smapL4SMaup,       1, MPI_LOGICAL,        0,MPICOMM,mpierr)
     call MPI_BCAST(N_obsbias_max,         1, MPI_INTEGER,        0,MPICOMM,mpierr)
    
-
-    ! create LandAssimAlarm
-
-    ! RR20210506: If AssimTime is indeed the time of the first land analysis, this should be "rrTime=AssimTime-ModelTimeStep"
-
-    !  NOT SURE THE ABOVE COMMENT MADE SENSE AT THE TIME OR NOW. NEED TO DOUBLE-CHECK AND CLEAN UP.
-    
-    rrTime = AssimTime - ModelTimeStep        ! probably rrTime=AssimTime is correct????
-
-    LandAssimAlarm = ESMF_AlarmCreate(                                          &
-         clock,                                                                 &
-         name='LandAssim',                                                      &
-         ringTime=rrTime,                                                       &
-         ringInterval=LandAssim_DT,                                             &
-         ringTimeStepCount=1,                                                   &
-         sticky=.false.,                                                        &
-         rc=status                                                              &
-         )
-    _VERIFY(status)
-
     !----
  
     if (.not. root_proc)  allocate(obs_param(N_obs_param)) 
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -1043,7 +1043,7 @@ contains
     type(ESMF_Time)              :: CurrentTime
     type(ESMF_Time)              :: AssimTime
     type(ESMF_Alarm)             :: LandAssimAlarm
-    type(ESMF_TimeInterval)      :: LandAssim_DT, LandAssim_OffDT
+    type(ESMF_TimeInterval)      :: LandAssim_DT, one_day
     integer                      :: LandAssimDTstep
     type(ESMF_TimeInterval)      :: ModelTimeStep 
     type(ESMF_Time)              :: rrTime, pertSeedTime
@@ -1076,7 +1076,7 @@ contains
     character(len=300)   :: fname_tpl
     character(len=14)    :: datestamp
     character(len=ESMF_MAXSTR) :: id_string 
-    integer              :: nymd, nhms, yy, dd, mm, h, m, s, Landassim_offset
+    integer              :: nymd, nhms, yy, dd, mm, h, m, s, Landassim_reftime, ref_time(6)
 
     !! from LDASsa
     
@@ -1155,17 +1155,31 @@ contains
     ! How many seconds later to start first land ananlysis, default is 3h = LandAssimDtStep
     call MAPL_GetResource(                                                      &
          MAPL,                                                                  &
-         landassim_offset,                                                     &
-         'LANDASSIM_OFFSET:',                                                  &
-         default= LandAssimDtStep,                                                       &
+         landassim_reftime,                                                     &
+         'LANDASSIM_REFTIME:',                                                  &
+         default= 000000,                                                      &
          rc=status                                                              &
          )
     _VERIFY(status)
 
-    _ASSERT(mod(landassim_offset, model_dtstep)==0,"inconsistent inputs for HEARTBEAT_DT and LANDASSIM_OFFSET")
+    s = MAPL_nsecf(landassim_reftime) 
+    _ASSERT(mod(s, model_dtstep)==0,"inconsistent inputs for HEARTBEAT_DT and LANDASSIM_REFTIME")
 
-    call ESMF_TimeIntervalSet(LandAssim_OffDT, s=landassim_offset, rc=status)
-    _VERIFY(status)
+    call ESMF_TimeGet(CurrentTime, YY = ref_time(1), &
+                                   MM = ref_time(2), &
+                                   DD = ref_time(3), &
+                                   rc=status) 
+    ref_time(4) = landassim_reftime/10000
+    ref_time(5) = mod(landassim_reftime,10000)/100
+    ref_time(6) = mod(landassim_reftime,100)
+
+    call ESMF_TimeSet( AssimTime, YY = REF_TIME(1), &
+                                  MM = REF_TIME(2), &
+                                  DD = REF_TIME(3), &
+                                  H  = REF_TIME(4), &
+                                  M  = REF_TIME(5), &
+                                  S  = REF_TIME(6), rc=rc )
+
 
     ! determine date and time of first land analysis   [R20210506: not sure I understand the intent here correctly.] 
     !
@@ -1182,7 +1196,15 @@ contains
     !   ..., [some date] 0z, 3z, 6z, 9z, ..., 21z, [next day], 0z, 3z, ... 21z, [another day later] 0z, 3z, ... 21z, [yet another day later] 0z, 3z, ...
     ! What's needed here is to determine AssimTime such that it the earliest time in the infinite sequence that is after the restart time.
 
-    AssimTime = CurrentTime + LandAssim_OffDT
+    if (AssimTime > CurrentTime) then ! going back one day
+       call ESMF_TimeIntervalSet(one_day, d=1, rc=status)
+       _VERIFY(status)
+       AssimTime = AssimTime - one_day
+    endif
+!
+    if (AssimTime <= CurrentTime ) then
+      AssimTime = AssimTime + (INT((CurrentTime - AssimTime)/LandAssim_DT)+1)*LandAssim_DT
+    endif
 
     call ESMF_UserCompGetInternalState(gc, 'TILE_COORD', tcwrap, status)
     _VERIFY(status)
@@ -1228,7 +1250,6 @@ contains
        enddo
     endif
     call MPI_Bcast(pert_rseed, NRANDSEED*NUM_ENSEMBLE, MPI_INTEGER, 0, mpicomm, mpierr)
-    
     
     allocate(N_catl_vec(numprocs))
     allocate(low_ind(numprocs))
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -1035,7 +1035,7 @@ contains
     type(ESMF_Clock),    intent(inout) :: clock  ! The clock
     integer, optional,   intent(  out) :: rc     ! Error code
     
-    integer :: status
+    integer                      :: status
     character(len=ESMF_MAXSTR)   :: Iam
     character(len=ESMF_MAXSTR)   :: comp_name
     
@@ -1044,7 +1044,7 @@ contains
     type(ESMF_Time)              :: AssimTime
     type(ESMF_Alarm)             :: LandAssimAlarm
     type(ESMF_TimeInterval)      :: LandAssim_DT, one_day
-    integer                      :: LandAssimDTstep
+    integer                      :: LandAssim_T0, LandAssimDTstep
     type(ESMF_TimeInterval)      :: ModelTimeStep 
     type(ESMF_Time)              :: rrTime, pertSeedTime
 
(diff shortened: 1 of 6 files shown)


gmao-rreichle:
@weiyuan-jiang 
I added one commit with minor cleanup but also a few questions
[5df0d1b](https://github.com/GEOS-ESM/GEOSldas/pull/397/commits/5df0d1b6c2c0b22950b561b4e1f73c161008c712)
In a nutshell, I'm not sure I understand how  exactly the branch worked w.r.t. the new "reftime".  I put my questions directly into the code.  We can clean up the comments later.  Let me know if you have questions, thanks. 
cc: @saraqzhang

saraqzhang:
@weiyuan-jiang :   what value is in AssimTime  at the place before creating LandAssimAlarm ?  If Landassim_DT is 10800, and we wanted assim alarm to ring at 5400, will AssimTIme be a negative number?  ( rrTime = AssimTIme+Landassim_DT-ModelStep) 
@gmao-rreichle

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1146,11 +1146,62 @@ subroutine Initialize(gc, import, export, clock, rc)
...
+    !             That is, the reftime should not need to change when the initial (restart) time changes.
+    !             e.g., reftime=0z,    initial time=0z, landassim_dt=3h  --> first analysis is at 3z    (next at 6z, 9z, ...)
+    !             e.g., reftime=3z,    initial time=0z, landassim_dt=3h  --> first analysis is at 3z
+    !             e.g., reftime=0z,    initial time=3z, landassim_dt=3h  --> first analysis is at 6z
+    !             e.g., reftime=1:30z, initial time=0z, landassim_dt=3h  --> first analysis is at 1:30z (next at 4:30z, ...)
+    !             e.g., reftime=7:30z, initial time=0z, landassim_dt=3h  --> first analysis is at 1:30z (next at 4:30z, ...)
+    ! If AssimTime is indeed meant to be the time of the first land analysis, I'm pretty sure hour/min/sec of AssimTime
+    !   would need to depend on "current" time (ie, "datestamp") as well as "reftime"
+    ! Put differently, "landassim_reftime" and "landassim_dtstep" effectively define an infinite sequence of land analysis time, something like:
+    !   ..., [some date] 0z, 3z, 6z, 9z, ..., 21z, [next day], 0z, 3z, ... 21z, [another day later] 0z, 3z, ... 21z, [yet another day later] 0z, 3z, ...
```
weiyuan-jiang, line 1183: Ok, my ref time is absolute time. Yours is relative to start time. I am fine with your definition

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1164,11 +1215,7 @@ subroutine Initialize(gc, import, export, clock, rc)
        _VERIFY(status)
        call MAPL_GetResource ( MAPL, fname_tpl, Label="LANDASSIM_OBSPERTRSEED_RESTART_FILE:", DEFAULT="../input/restart/landassim_obspertrseed%s_rst", RC=STATUS)
        _VERIFY(STATUS)
-       call MAPL_DateStampGet( clock, datestamp, rc=status)
-       _VERIFY(STATUS)
-       read(datestamp(1:8),*)   nymd
-       read(datestamp(10:13),*) nhms
-       nhms = nhms*100
+       nhms = landassim_reftime  ! RR20210506: Why is this not hour/min/sec from AssimTime (ie, the time of the first land analysis)?
```
weiyuan-jiang, line 1218: This is absolute time for the output of rseed restart file

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1277,13 +1320,11 @@ subroutine Initialize(gc, import, export, clock, rc)
...
 
-    ! create LandAssimAlarm with option of centered update 
+    ! create LandAssimAlarm
 
-    if (centered_update)then
-       rrTime = CurrentTime+Landassim_DT/2-ModelTimeStep 
-    else
-       rrTime = CurrentTime+Landassim_DT-ModelTimeStep 
-    endif
+    ! RR20210506: If AssimTime is indeed the time of the first land analysis, this should be "rrTime=AssimTime-ModelTimeStep"
```
weiyuan-jiang, line 1325: We don't do the analysis at the very beginning. Do we?

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: If it is relative to start time. Some of my calculation needs to be changed

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
saraqzhang: it has not  been done in ldas.

gmao-rreichle: No, there is no analysis at the restart time.  This may be useful to have in the future.  For now, however, we can assume that the first analysis has to be after the first model_dtstep but no later than after 24 hours.

weiyuan-jiang: Before I do further changes, I want to make sure what ref_time means. For example, if the program starts at 6z , and does analysis at 7:30, according to your definition, ref_time = 1:30, according to mine, it is 7:30.  We only need one definition. @gmao-rreichle

saraqzhang:
correction:  Assimalarm = CurrentTime + refTime - ModelTimeStep

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
gmao-rreichle: The ref_time cannot depend on the restart time.  It has to be more general.  Two variables, ref_time and landassim_dt, determine an infinite sequence of land analysis times (see my comments in the code).  In the example above, the user must be able to specify either ref_time=1:30 **or** ref_time=7:30 **or** ref_time=22:30, and the effect must always be that a land analysis happens whenever the clock hits 1:30, 4:30, 7:30, 10:30, ..., 22:30 on any day.   Otherwise, the user would have to change the ref_time whenever they restart the program from a different time-of-day.
I suppose a modulus operation on the ref_time at initialization ( mod(ref_time_in_seconds/dtstep_in_seconds) ) could be used to have the ref_time always be the earliest possible time-of-day at which a land analysis should occur for the purpose of standardizing calculations within the code.  Again, though, specifying 0<=ref_time+N*landassim_dt<=23:59:59 for any N (within the 0...23:59:59 limits) must results in the same outcome.

gmao-rreichle:
Sorry, I accidentally closed the Issue.  @saraqzhang : see my comment in response to @weiyuan-jiang above.  There cannot be a "preferred" ref_time.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1146,10 +1146,51 @@ subroutine Initialize(gc, import, export, clock, rc)
...
+    call MAPL_GetResource(                                                      &
+         MAPL,                                                                  &
+         landassim_reftime,                                                     &
+         'LANDASSIM_REFTIME:',                                                  &
+         default= hhmmss,                                                       &
+         rc=status                                                              &
+         )
+    _VERIFY(status)
+
+    ref_seconds = landassim_reftime/10000*3600 +mod(landassim_reftime,10000)/100 *60 + mod(landassim_reftime,100)
```
gmao-rreichle, line 1170: I might not be reading this correctly, but I think ref_seconds is simply the ref_time converted into seconds.  To be safer and clearer, I think we should add parentheses in the "minutes" term, as in:
`+(mod(landassim_reftime,10000)/100) *60
`
But that's not the real issue.  If the above is correct, then a ref_time of 073000 will result in 27,000 seconds.  This interval is not what we're looking for.
I think what we need is the following:
dt_ref = mod( ref_seconds, LandAssimDtStep)
For LandAssimDtStep=10800 (3 hours), dt_ref will then be 5400 seconds for **all** of the following ref_times: 013000, 043000, 073000, ..., 223000.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1146,10 +1146,51 @@ subroutine Initialize(gc, import, export, clock, rc)
...
     _VERIFY(status)
+    
+    _ASSERT(mod(LandAssimDtStep,model_dtstep)==0, "inconsistent inputs for HEARTBEAT_DT and LANDASSIM_DTSTEP")
+
+    call ESMF_TimeIntervalGet(LandAssim_DT, H = h, m = m, s = s, rc=status)
+    _VERIFY(status)
+
+    ! read reference time for land analysis from resource file (default: LandAssim_DT=3z)
+    ! ref_time is hhmmss format
+    hhmmss = h*10000 + m*100 + s
```
gmao-rreichle, line 1160: The default for ref_time should simply be 000000 (ie, 0)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: @gmao-rreichle , The ref_time defined above makes things a little bit too complicated.  If we defined ref_time as a number between 0 and assim_dtstep and the assim_time = currentTime + ref_time + n * assime_dtstep ( n=0,1,2,......) , that would be easier.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: Here ref_time is in hhmmss format and the maximum seconds is LandAssimDtStep

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: Again, my definition of the ref_time is a reference time of the program starting time. It should be in hhmmss format but the total seconds is less than LandassimDTstep. I am open to different definitions

gmao-rreichle: I think I see what you're trying to do.  That is, limit the allowable rc input options for ref_time input to fall within the interval [0, LandAssimDtStep].  Put differently, users need to pick a ref_time that assumes a 0z restart and then specify the first AssimTime for the 0z restart case. 
I think this is too complicated to explain in the rc file.  If a user restarts at 12z and wants a land analysis at 13:30z, they're going to have a hard time figuring out that they need to specify 01:30z for the ref_time.   I still think we need to do the modulus calc within the src code to allow the user so specify any ref_time that's spaced LandAssimDtStep apart from the desired smallest AssimTime_of_day.   I think it's not prohibitively difficult to implement.  The pseudocode in my earlier comment should work.

saraqzhang:
Look at the original code of create alarm clock,  how it did the normal ringing work:   "ringTime = CurrentTime+Landassim_DT-ModelTimeStep"   makes the first ring  (here it is 3h after  whenever the run starts).  After that  "ringInterval = Landassim_DT " controls  the subsequent ring times ( here is every 3h) .   
The previous test runs I did with  " ringTime = CurrentTime +  Landassim_DT/2 -ModelTimeStep "  and  "ringInterval = Landassim_DT " carried out analyses at "centered" time,  it doesn't matter when the restart time is.  I did restart time at 21z, the analyses are at 22:30, 1:30,.... I did restart time at 09z, the analyses are at 10:30, 13:30 ....  
 @weiyuan-jiang @gmao-rreichle

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: The Offset time is 5400 s and doesn't need to be in the format of hhmmss

weiyuan-jiang:
@saraqzhang , I think the code is right. But the name "ref_time" will confuse users and ourselves.

weiyuan-jiang:
After renaming LANDASSIM_reftime to LANDASSIM_OFFSET,  I think the program is cleaner and clearer

gmao-rreichle:
Hi @weiyuan-jiang 

If we stick with your concept of specifying the "offset" time, then the new variable name is indeed clearer.  However, the original motivation for "ref_time" was based on the "ref_time" concept in MAPL HISTORY.   It's possible that I misunderstand the MAPL HISTORY ref_time concept.  In this the case, somebody would have to explain the HISTORY "ref_time" to me again.  

But if my understanding of the HISTORY "ref_time" concept is correct, then I don't see why we wouldn't adopt it for configuring the land analysis times.  It would be easier to explain to users.  The HISTORY "ref_time" approach also seems more robust in that users don't have to make any connection whatsoever between the restart time and the land analysis times.  They can specify any of 86400/landassim_dtstep options (ie, 013000, 043000, 073000, ... 223000 all have the same effect).
  
Come to think of it, we should really use LANDANALYSIS_FREQUENCY (instead of dtstep) and LANDANALYSIS_REFTIME in the RC file to match the HISTORY language.  (As an aside, this would then also be used to verify the consistency of the "frequency" and "reftime" settings of the "catch_progn_incr" and "lndfcstana" HISTORY collections.
Perhaps it's best to talk this through at our regular tag-up tomorrow.  I think we've exhausted what we can accomplish via email and Github comments.

gmao-rreichle:
> it is cleaner. We can instruct users that if offset time in rc is set to certain number, the refTime for analysis outputs in History.rc should be modified accordingly.
> should I pull the version and run a test.

@saraqzhang : Please hold off, see my comment above (~noon today)

weiyuan-jiang:
After I take more close look at RefTime in History, I figured out how it works. 1) if refime = current time, alarm rings at reftime+ assimdt_step 2) if ref_time < current time, it will increase by multiples of assimdtstep until it is bigger than current time.  3)  if reftime > current time,  the alarm rings until ref_time.  So if it starts at 0,  reftime is 9:00,  there is no alarm in 3, 6.   This is different than we thought and is not what we want. @gmao-rreichle

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1277,26 +1344,26 @@ subroutine Initialize(gc, import, export, clock, rc)
...
-    if (centered_update)then
-       rrTime = CurrentTime+Landassim_DT/2-ModelTimeStep 
-    else
-       rrTime = CurrentTime+Landassim_DT-ModelTimeStep 
-    endif
+    ! RR20210506: If AssimTime is indeed the time of the first land analysis, this should be "rrTime=AssimTime-ModelTimeStep"
+
+    !  NOT SURE THE ABOVE COMMENT MADE SENSE AT THE TIME OR NOW. NEED TO DOUBLE-CHECK AND CLEAN UP.
+    
+    rrTime = AssimTime - ModelTimeStep        ! probably rrTime=AssimTime is correct????
```
gmao-rreichle, line 1353: @weiyuan-jiang, I'm not sure this is correct.  Shouldn't it be rrtime=AssimTime after the recent revisions?

saraqzhang:
I suggested to move up the alarm clock  to be right after the AssimTime calculation.  ( line 1355--1364   move up to line 1214).

gmao-rreichle:
> I suggested to move up the alarm clock to be right after the AssimTime calculation. ( line 1355--1364 move up to line 1214).

Sorry, @saraqzhang, I misread your message. Should have stopped working a bit earlier... 
I moved the alarm:  [7ca046b](https://github.com/GEOS-ESM/GEOSldas/pull/397/commits/7ca046b6b8a7287288449b5685ae3a009c7b3cf1)
Note that this commit also tentatively addressed my question whether ringTime should be AssimTime or AssimTime-model_dtstep (see my now-outdated inline comment).
cc: @weiyuan-jiang

saraqzhang:
I think there is some reason why the ringtime was set at  AssimTime - ModelTimeStep, maybe something to do with before or after the  model timestep ticking. 
see in AssimGC  RUN ()   :   
call esmf2ldas(ModelTimeCur+ModelTimeStep, date_time_new, rc=status) 
...
call get_enkf_increments(  date_time_new, ...                                    )

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: The CurrentTime is not updated until HistoryGC run().  So when is time to do data assimilation, the time in LandAssimGc is still AssimTim -  ModelTimeStep and it should ring at that time

gmao-rreichle (approved):
Looks like we have everything done.
I'll merge, this will then bring the changes into Sara's original PR #[392](https://github.com/GEOS-ESM/GEOSldas/pull/392)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
gmao-rreichle: Fixed with most recent commit.

saraqzhang:
@weiyuan-jiang @gmao-rreichle 
Will you make update to ldas_setup for the changes here?   Though not related to this PR, could we use the opportunity to add one line to line number 1490 :  
print 'ADAS_EXPDIR:  '
so to include this parameter in the sample txt file  when user runs ldas_setup  sample.

gmao-rreichle:
@saraqzhang : This could be included in PR #392, but perhaps it's better to create a new branch & PR.  I'm hoping we can merge #392 within the next couple of days, if Biljana's testing and your verification are successful, and then move quickly to a new Release.  The ADAS_EXPDIR changes may or may not get done as quickly.

PR #413, opened 2021-05-27
Add CircleCI to GEOSldas

### LINKED ISSUE #285: Add CircleCI Tests
A question mainly for @gmao-rreichle. Would you like me to add CircleCI build tests (on PRs a la GEOSgcm, etc.) to GEOSldas? I think it would be fairly simple, and now that we have access to bigger resources there, build tests will probably be sub-10 minutes.

My only need would be to know exactly what you want checked out. For example, on GEOSgcm we currently do:
```
mepo clone
mepo develop GEOSgcm_App GEOSgcm_GridComp
```
I'm guessing GEOSgcm_App isn't needed, but I can change to any develop branch. I could also do like GEOSgcm_GridComp where we have:
```
mepo checkout-if-exists ${CIRCLE_BRANCH}
```
where if someone has a branch, say, here and in GEOSgcm_GridComp with the same name, we can grab both at the same time.

mathomp4:
I was looking at my Issues on GitHub and saw #285 and realized, huh, I never did anything on that. So...

Closes #285 

This is my draft attempt to add CircleCI tests on GEOSldas. This means any PR on this repo will need to be able to compile to be accepted.

Let's see if it even works first.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,7 +43,7 @@ jobs:
           command: |
             cd ${CIRCLE_WORKING_DIRECTORY}/GEOSldas
             echo "${CIRCLE_BRANCH}"
-            if [ "${CIRCLE_BRANCH}" != "develop" ] && [ "${CIRCLE_BRANCH}" != "master" ] && [ "${CIRCLE_BRANCH}" != "main" ]
+            if [ "${CIRCLE_BRANCH}" != "develop" ] && [ "${CIRCLE_BRANCH}" != "main" ]
             then
                mepo checkout-if-exists ${CIRCLE_BRANCH}
             fi
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,63 @@
+version: 2.1
+
+executors:
+  gcc-build-env:
+    docker:
+      - image: gmao/ubuntu20-geos-env-mkl:v6.1.0-openmpi_4.0.5-gcc_10.2.0
+        auth:
+          username: $DOCKERHUB_USER
+          password: $DOCKERHUB_AUTH_TOKEN
+    environment:
+      OMPI_ALLOW_RUN_AS_ROOT: 1
+      OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
+      OMPI_MCA_btl_vader_single_copy_mechanism: none
+    resource_class: large
+    #MEDIUM# resource_class: medium
+
+workflows:
+  version: 2.1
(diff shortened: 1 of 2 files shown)


mathomp4:
Aaaand...nothing was triggered. Sigh. Documentation time for Matt!

mathomp4:
Okay. So this is broken at the moment for some MKL reason. I'll close this and do some tests by hand...

mathomp4:
Okay. I think I figured it out. Honestly, I'm not sure how GEOSldas has been building as it seems to be rather explicitly missing something in CMake.

mathomp4:
Yay. 🎉

gmao-rreichle (approved):
Thanks, Matt.  I think this will be helpful!

mathomp4:
@gmao-rreichle Note: I'm also looking at maybe adding a GEOSldas build test to some of the other repos as well once this seems "good". I'll have to ponder the best way to do it as your `main` is a bit "longer lasting" than GEOSgcm. We usually just build against the `main` of GEOSgcm, but I'd probably want to go against `develop` of GEOSldas when testing, say, GEOSgcm_GridComp PRs...

bena-nasa (approved):

gmao-rreichle:
> @gmao-rreichle Note: I'm also looking at maybe adding a GEOSldas build test to some of the other repos as well once this seems "good". I'll have to ponder the best way to do it as your `main` is a bit "longer lasting" than GEOSgcm. We usually just build against the `main` of GEOSgcm, but I'd probably want to go against `develop` of GEOSldas when testing, say, GEOSgcm_GridComp PRs...

@mathomp4, thanks for looking out for GEOSldas.  If I understand correctly, you would require, say, GEOSgcm_GridComp PRs to pass a GEOSldas build test, correct?  In this case, yes, you would need to build GEOSldas "develop".  The GEOSldas "main" branch does not necessarily work with GEOSgcm_GridComp "develop".
One more thing: It makes sense to **require** that a PR for GEOSgcm_GridComp "develop" compiles with GEOSldas "develop". But what if someone has a PR that is for another branch?  We don't want to block a merge that isn't going into GEOSgcm_GridComp "develop".   Could CircleCI be set up to only kick in for merging into "develop"?  Also, would this impact Scott merging GEOSgcm_GridComp "develop" into "main"?  I can't quite wrap my mind around the latter scenario, and I hope that's a non-issue.
Let me know if this requires more discussion.  Thanks again, R
cc: @sdrabenh @weiyuan-jiang @biljanaorescanin

PR #416, opened 2021-06-17
Update to Baselibs 6.2.4. Enable GFE Namespace

mathomp4:
This PR brings GEOSldas up-to-date with GEOSgcm v10.19.2. The change is two-fold:

1. Updates to Baselibs 6.2.4
2. Enables GFE Namespace in CMake

Note that the nightly testing framework will need to change with it so that the GNU tests are using the 6.2.4 Baselibs g5_modules (if not, build failure).

Note 2: I'm marking as 0-diff as it should be, but should be tested!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 executors:
   gcc-build-env:
     docker:
-      - image: gmao/ubuntu20-geos-env-mkl:v6.1.1-openmpi_4.0.5-gcc_10.3.0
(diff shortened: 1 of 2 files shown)


gmao-rreichle:
Ok with me as long as it's 0-diff.
@biljanaorescanin, can you please test this? Note @mathomp4's comment about needing to modify the GNU tests for this to work.

biljanaorescanin:
@gmao-rreichle This PR was tested in 2 parts. First part tested PR as it was with no additional changes and in that case intel tests passed and GNU tests failed for build (as we expected). For second part Matt changed baselibs we use,  I retested PR and all is passing now.

gmao-rreichle (approved):

PR #418, opened 2021-06-24
parallel fft within nodes

weiyuan-jiang:
1. Before this branch, each processor computes full domain's fft independently without communication or data exchange.  In this branch, each node computes full domain's fft without data exchange among nodes.  However, all the processors in a node use shared memory to compute fft parallelly.  The full domain is divided among the processors evenly and each processor calculate part of the domain.  This change is **NOT zero-diff!**
2. The modifications also refine the assignment of tiles to processors (better than previous version but not quite as optimal as the "painter's partition" algorithm).
3. To ensure 0-diff across processor types (architecture), limit MKL's freedom to choose algorithms during runtime (MKL_CBWR=AVX2).  This achieves 0-diff across haswell and skylake nodes for the assimilation test case in the EASE grid tile space.  (Note that this does not resolve the layout dependency of the assimilation test case in the cube-sphere tile space.)


weiyuan-jiang:
By default, as long as there is mkl library, it will perform parallel fft, which is non-zero diff. Should we introduce another control parameter to pick serial of parallel?  I think it is not necessary if we fully test this branch and verify the non-zero diff is acceptable. @gmao-rreichle

gmao-rreichle:
> By default, as long as there is mkl library, it will perform parallel fft

That is, if MKL is not available, the code will use serial fft.  Is this correct?

> Should we introduce another control parameter to pick serial of parallel?

I suppose this would be needed if each processor has its own memory (ie, 1 processor per node), which we can assume is not going to be the case at NCCS for the foreseeable future, although I'm not sure what external collaborators are using.  So for now at least we don't need the extra control parameter.

Also, I did not see AVX2 in the "Files changed" tab to ensure 0-diff across hasw and skylake.  Don't we need to add the MKL constraint somewhere in setup?

src/Applications/LDAS_App/preprocess_ldas_routines.F90
```diff
@@ -2152,32 +2153,53 @@ subroutine optimize_latlon(fname,arg)
...
+            endif
           enddo
           local = j
        enddo
-       
+      if( sum(JMS) /= JMGLOB) then
+          print*, sum(JMS), JMGLOB
+          stop ("wrong cs-domain distribution in the first place")
+       endif 
        ! adjust JMS.rc make make sure no process has 0 or 1
```
gmao-rreichle, line 2203: What happens if a user (inadvertently) requests more processors than tiles, or if they are running only a single (cube-sphere) tile on one processor?  In this case, there would have to be at least one processor with 0 or 1 tiles.  
I don't think anyone has run a single cube-sphere tile yet, but users occasionally run a single  EASE grid tile.
Unless I misunderstand the comments (and I didn't manage to follow the algorithm implementation yet for lack of time), we might need to allow the case of a single cubesphere tile on a single processor and also provide a more helpful stop message if the user requests more processors than there are tiles.

src/Applications/LDAS_App/preprocess_ldas_routines.F90
```diff
@@ -2152,32 +2153,53 @@ subroutine optimize_latlon(fname,arg)
           n2 = k*IMGLOB
           
           do i=1,60
-             rates(i) = (i-1)*0.1
+             rates(i) = -0.3 + i*0.01
```
gmao-rreichle, line 2156: Are these changes related to the "painter's partition" algorithm?  If so, we should probably add a comment somewhere that the distribution implemented here follows this algorithm.

src/Applications/LDAS_App/preprocess_ldas_routines.F90
weiyuan-jiang: To run cubed-sphere grid, the minimum number of processors is 6. Here the comment  "adjust JMS.rc make make sure no process has 0 or 1" is about the number of grid cells, not tiles.

src/Applications/LDAS_App/preprocess_ldas_routines.F90
weiyuan-jiang: This is another effort to optimize the distribution. It is supposed to be better but not really as optimal as painter's partition algorithm.

src/Applications/LDAS_App/preprocess_ldas_routines.F90
gmao-rreichle: Ok, thanks, but I still don't understand exactly what is and isn't possible, and what the stop/exit messages are when something doesn't work.  Let me ask the question differently:
If a user wants to run a single tile from a cube-sphere tile space, do they have to request 6 processors?  
If a user wants to run a single tile from a cube-sphere tile space, would the program stop because for a single-tile domain there can never be at least two tiles on each processor (regardless of how many processors are requested)?
Sorry for being slow to catch on.

weiyuan-jiang: This is sanity check. For example, if it is c72, after distribution, the sum of grid cells is not 72, then it stops.  1) yes, users need to have 6, 12, 18, ..... processors. Actually this is not our requirement, it is MAPL's requirement.  2) Again this is about grid cells, not tiles. For example, if it is c72, the number of processor for a face should not exceed 36 ( we divided the grid along one dimension). If it is bigger than 36, then some processors would end up with 1 grid cell.

weiyuan-jiang: It is fine to run with one tile and 6 processors

weiyuan-jiang: Corrected: If it is c72, after distribution, the sum of grid cells is not 6*72, the it stops

gmao-rreichle: Thanks, @weiyuan-jiang.  I tried my luck at adding some help text about the domain decomposition in this latest commit:
[0d376a0](https://github.com/GEOS-ESM/GEOSldas/pull/418/commits/0d376a083a8de82c1cbbb5e641976ab09c1bea27?w=1)
The commit also contains minor cleanup (incl. added "intent" statements)

gmao-rreichle:
I merged the MKL_CBWR=AVX2 change from @weiyuan-jiang's separate branch into this branch ("parallel fft with shared memory") since the parallel FFT heavily draws on MKL.

gmao-rreichle (approved):
I successfully ran the standard suite of "nightly" tests, which resulted in the expected failure of the assimilation test cases (GLOBAL/assim, GLOBALCS/assim, GNUGLOBAL/assim, and GNUGLOBALCS/assim).
I verified that the ObsFcstAna files and LandPert restarts are different only by roundoff.  
The above-mentioned tests need to be re-baselined after this PR is merged.

PR #419, opened 2021-06-27
Use RTM parameters from SMAP L2 retrievals 

gmao-rreichle:
Addresses changes that are needed to use L-band radiative transfer model (RTM) parameters derived from the SMAP L2 passive retrieval product.


gmao-rreichle:
@gmao-qliu,  I created this pull request, merged the latest develop into the branch, and added one commit with some minor cleanup and comments for discussion at our next meeting.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -2592,13 +2602,28 @@ subroutine get_mwrtm_param(internal,N_catl, rc)
...
+    real, dimension(:), pointer :: DCATAU   ! needed?? declared as global var in module!?  reichle, 27Jun2021
     
     integer :: N_catl_tmp, n, mpierr, status
     logical :: mwp_nodata, all_nodata_l
-    
+
+    character(len=ESMF_MAXSTR)  :: TAUFile
+    type(ESMF_Time)             :: CURRENT_TIME
+ 
     if(allocated(mwRTM_param)) then
```
gmao-rreichle, line 2613: @weiyuan-jiang, @gmao-qliu,
I'm grinding my way through the revisions of this PR and just stumbled across this "if" statement, which remains unchanged in Qing's branch.  If I understand the "if" statement here correctly, it looks like Qing's branch runs "get_mwrtm_param()" and thus reads DCATAU from the file ONLY ONCE per segment.  After "get_mwrtm_param()" has been allocated and filled, the DCATAU file is ignored.  Do I understand this correctly?

gmao-rreichle: The above understanding is consistent with the original use case for get_mwrtm_param() and also makes sense in that the variable mwRTM_param is allocated but never deallocated.
If this is all true, we need to fix the code and then re-run all the science tests from scratch, which would mean that this PR is no longer a candidate for the Version 6 L4_SM system. 
But maybe I'm missing something.  It's been a long day already.

gmao-rreichle:
continued in PR #521

PR #424, opened 2021-07-13
bug fix in no-data check in mwRTM_routines.F90

gmao-rreichle:
This bug was present from the very first implementation of the microwave radiative transfer model (mwRTM) in the LDASsa CVS tag `reichle-LDASsa_m2-10`.
I'm 99.99% confident that the fix is 0-diff, but I haven't tested it.
@biljanaorescanin, can you please test?  If it's indeed 0-diff, go ahead and merge. 

cc: @gmao-qliu

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/mwRTM_routines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/mwRTM_routines.F90
@@ -369,9 +369,9 @@ contains
        ! to check one field of mwp because mwRTM_param_nodata_check()
        ! is called in mwRTM_get_param()]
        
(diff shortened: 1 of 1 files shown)


biljanaorescanin:
GLOBAL assim test case failed, due to the slurm error, runtime limit,  it stopped during writing of **landpert_internal_rst** file. 
GNUGLOBAL assim test failed with the same error.

gmao-rreichle:
Thanks, @biljanaorescanin.  It looks like something is wrong with Discover.  I highly doubt that the test failures are related to this PR.  Please re-run the tests when Discover is back to normal.

biljanaorescanin:
I've rerun the tests, all passed.

biljanaorescanin (approved):

PR #425, opened 2021-07-14
Fix use of mwRTM parameters from L2_SM_P_E files

gmao-rreichle:
@gmao-qliu, @weiyuan-jiang,
I committed my changes to Qing's branch on a separate branch to make it easier to compare to Qing's original version. 
There is still a lot of work needed to make this branch ready for a repeat of the science development runs, see below. 
-- Rolf

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -814,6 +814,17 @@ contains
        RC=STATUS  ) 
   VERIFY_(STATUS) 
   
+  ! Exports for microwave radiative transfer model (mwRTM)
+
+  call MAPL_AddExportSpec(GC                                                      ,&
+       LONG_NAME          = 'L-band Microwave RTM: Vegetation opacity (normalized with cosine of incidence angle)'   ,&
+       UNITS              = '1'                                                   ,&
+       SHORT_NAME         = 'MWRTM_VEGOPACITY'                                    ,&
+       DIMS               = MAPL_DimsTileOnly                                     ,&
+       VLOCATION          = MAPL_VLocationNone                                    ,&
+       RC=STATUS  )
+  _VERIFY(STATUS)
+
   !
   ! INTERNAL STATE
   !
@@ -1480,6 +1491,10 @@ contains
     real, dimension(:),pointer :: PRMC_ana=>null()     ! profile soil moisture
     real, dimension(:),pointer :: TPSURF_ana=>null()   ! tpsurf
     real, dimension(:),pointer :: TSOIL1_ana=>null()   ! tsoil1
+
+    !! export for microwave radiative transfer model (mwRTM)
+
+    real, dimension(:),pointer :: MWRTM_VEGOPACITY=>null()  ! vegetation opacity (time-varying)
     
     logical, save              :: firsttime=.true.
     type(cat_bias_param_type)  :: cat_bias_param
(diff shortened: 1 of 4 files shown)


gmao-rreichle:
The first commit includes the following (not yet tested!):
- fixed no-data issues associated with use of MAPL_ReadForcing
- fixed no-data handling in if statement for computing Tb
- fixed lack of backward compatibility with use of  current (calibrated) mwRTM parameter file
- renamed DCATAU and TAU to VEGOPACITY
- added export spec for VEGOPACITY
- cleaned up MPI types

**NOT YET ADDRESSED:**
- read VEGOPACITY from file when needed
- add VEGOPACITY to GEOSldas_HIST.rc, tile_bin2nc4.F90
- add VEGOPACITY_FILE to ldas_setup, GEOSldas_LDAS.rc
- mwrtm_bin2nc4.F90 and other pre-processing tools (?)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -2602,28 +2630,48 @@ subroutine get_mwrtm_param(MAPL, clock, N_catl, INTERNAL, rc)
...
     
     integer :: N_catl_tmp, n, mpierr, status
     logical :: mwp_nodata, all_nodata_l
 
-    character(len=ESMF_MAXSTR)  :: TAUFile
+    character(len=ESMF_MAXSTR)  :: VEGOPACITYFile
     type(ESMF_Time)             :: CURRENT_TIME
- 
+
+    ! it looks like this if statement prevents us from reading the vegetation opacity from the file after
```
weiyuan-jiang, line 2642: We probably need another subroutine get_vegopacity()

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -2602,28 +2630,48 @@ subroutine get_mwrtm_param(MAPL, clock, N_catl, INTERNAL, rc)
...
+    ! time-varying data in some other way.
+    ! - reichle 13 July 2021
     if(allocated(mwRTM_param)) then
        _RETURN(_SUCCESS)
     endif
 
-    allocate(DCATAU(N_catl))
-    call MAPL_GetResource(MAPL, TAUFile, label = 'TAU_FILE:', &
-        default = 'tau.dat', RC=STATUS )
+    allocate(VEGOPACITY(N_catl))
```
weiyuan-jiang, line 2652:
```suggestion
    allocate(VEGOPACITY(N_catl), source=MAPL_UNDEF)
```

weiyuan-jiang: it will be assigned to mwRTM structure if there is no file. So the default value here is necessary.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -2149,7 +2174,10 @@ subroutine CALC_LAND_TB(gc, import, export, clock, rc)
     real, dimension(:), pointer :: BH
     real, dimension(:), pointer :: BV
     real, dimension(:), pointer :: LEWT
-    real, dimension(:), pointer :: DCATAU 
+
+    ! "vegopacity" is NOT an internal state; do we need this pointer? if so, what does it do?
+    ! - reichle, 13 July 2021
+    real, dimension(:), pointer :: VEGOPACITY
```
weiyuan-jiang, line 2180: I think it can be removed safely here

gmao-rreichle:
Thanks, @weiyuan-jiang, for the comments.  I implemented the two easy ones and also fixed a few silly errors that prevented the branch from building successfully. The above "NOT YET ADDRESSED" list remains valid.

weiyuan-jiang:
Ok, I will take care of the "NOT Yet addressed" issues

weiyuan-jiang:
I will work on your branch

weiyuan-jiang:
@gmao-rreichle , where is ( or will be ) the file  VEGOPACITY located?

gmao-rreichle:
> @gmao-rreichle , where is ( or will be ) the file VEGOPACITY located?

The plan is to keep the file in the same directory as the mwRTM "bcs" file.  I suppose ldas_setup will have to create a link, something like `../input/vegopacity.data`

Come to think of it, we could make it such that we only specify an mwRTM **_directory_** name (much like for bcs), and then use standardized file names within the directory, that is:
```
/some_path/mwRTM_params_some_version/mwRTM_param.nc4
/some_path/mwRTM_params_some_version/vegopacity.bin
```
Then we wouldn't need two resource variables.  If the vegopacity file is not there (ie, the "old" way), we use the (bh,bv,lewt) info from mwRTM_param.nc4.  If the vegopacity is there, the two files must be constructed so that for a given tile, they either provide (bh, bv, lewt) or (vegopacity), which is what @gmao-qliu and I already agreed on.

Not sure if this idea is compatible with perhaps using an existing "landassim GridComp" (ie, mwRTM) restart file. 

We'll probably have to iterate a bit on the exact approach.  Since @gmao-qliu will need to re-do all the science experiments, we have plenty of time to fine-tune this.  The PR will not be merged into develop before September at the earliest.  What's needed now is a clean branch that works as intended (ie, reads vegopacity as needed and is backward compatible, ie, reproduces develop when given the "old" mwRTM_param file and no vegopacity file).

gmao-rreichle:
continued in PR https://github.com/GEOS-ESM/GEOSldas/pull/521

PR #426, opened 2021-07-14
add new subroutine get_vegopacity

weiyuan-jiang:
If we put subroutine get_vegopacity to the right location, it will work for testing purpose. @gmao-rreichle


gmao-rreichle:
@weiyuan-jiang, thanks for putting this together.  I've been staring at the code again this evening.  I think I know how to  rearrange get_mwRTM_param() to do what's needed, but I can't do anything else today.  I'm also not sure how much I can do tomorrow or Friday, because I'm already booked with meetings for most of the day.  I'm starting to think that we may not be able to get this ready before my leave.  Since it's no longer going into L4_SM Version 6, there's no rush, and in any case Qing has to put most of her effort towards Version 6.  
cc: @gmao-qliu

gmao-rreichle:
continued in PR https://github.com/GEOS-ESM/GEOSldas/pull/521

Issue #422, opened 2021-07-09
Implement MODIS snow cover fraction assimilation


gmao-rreichle:
@jpark50 is implementing the MODIS snow cover fraction assimilation of Toure et al. (2018), [doi:10.3390/rs10020316](https://doi.org/10.3390/rs10020316) in GEOSldas.

Ally's original implementation dates back to ~2012 and was never merged into the main development branch.
The original implementation was added on top of a very outdated version of LDASsa (`reichle-LDASsa_m2-09`, predating the earliest version mentioned in the GEOSldas [CHANGELOG](https://github.com/GEOS-ESM/GEOSldas/blob/main/doc/CHANGELOG.md).

A brief analysis of the differences and the location of the LDASsa source code is here:
**[readme_AllyToure_LDASsa_MODIS_SCF_assimilation_20210708.txt](https://github.com/GEOS-ESM/GEOSldas/files/6787864/readme_AllyToure_LDASsa_MODIS_SCF_assimilation_20210708.txt)**

cc: @lcandre2 @weiyuan-jiang @biljanaorescanin

PR #434, opened 2021-07-19
Add MODIS snow cover assimilation

jpark50:
Addresses #422.  The purpose of this PR is to track the any changes in LDAS with regards to the MODIS snow cover assimilation of Toure et al. (2018), doi:10.3390/rs10020316 in GEOSldas.


jpark50:
@lcandre2 @gmao-rreichle, I just finished debugging and it can be complied successfully (at my sandbox) and pushed a commit. I will check up the MODIS SCA reader in lis.

gmao-rreichle:
@jpark50, I'm making a first pass through the branch as it is right now.  I'll probably add a few "inline" comments along they way.
One quick thing:  Your branch includes the following binary file:
  `src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/.clsm_ensupd_upd_routines.F90.swo`
I don't know what this is, and how it came to be part of your branch, but I'm quite sure it shouldn't be.  You should be able to clean this up with "git rm ...", but please check the documentation before you use this command.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1367,6 +1374,46 @@ subroutine apply_enkf_increments( N_catd, N_ens, update_type, &
...
+            date_time%min  /= 0 .and. &
+            date_time%sec  /= 0         ) then
+
+          if (logit) write (logunit,*) 'no application of increments at this time'
+          cat_progn_has_changed = .false.
+          return
+
+       end if
+
+       cat_progn_has_changed = .true.     ! conservative initialization
```
gmao-rreichle, line 1391: The statement: `cat_progn_has_changed = .true. ! conservative initialization` does not make sense here.  This initialization is already at the beginning of the subroutine.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -1452,23 +1465,24 @@ subroutine get_obs_pred(                                       &
...
+         get_sfmc_lH, get_rzmc_lH, get_tsurf_lH, get_FT_lH, get_asnow_lH, get_Tb_lH, N_fields)
     
     ! allocate and assemble tile_data_l
-    
+    ! added get_asnow_lH at the function jpark50
     if (allocated(tile_data_l))  deallocate(tile_data_l)
     allocate(tile_data_l(N_catl,N_fields,N_ens))
     call get_obs_pred_comm_helper( N_catl, N_ens, N_TbuniqFreqAngRTMid,          &
-         get_sfmc_lH, get_rzmc_lH, get_tsurf_lH, get_FT_lH, get_Tb_lH, N_fields, &
+         get_sfmc_lH, get_rzmc_lH, get_tsurf_lH, get_FT_lH, get_asnow_lH, get_Tb_lH, N_fields, &
```
gmao-rreichle, line 1485: It looks like "asnow=asnow_l" still needs to be added to the list of optional input arguments in this call.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -1480,11 +1494,11 @@ subroutine get_obs_pred(                                       &
...
     
     ! read out sfmc, rzmc, etc. from tile_data_lH    
-    
+    ! addded asnow option at here
     call get_obs_pred_comm_helper( N_catlH, N_ens, N_TbuniqFreqAngRTMid,         &
-         get_sfmc_lH, get_rzmc_lH, get_tsurf_lH, get_FT_lH, get_Tb_lH, N_fields, &
+         get_sfmc_lH, get_rzmc_lH, get_tsurf_lH, get_FT_lH, get_asnow_lH, get_Tb_lH, N_fields, &
          option=2, tile_data=tile_data_lH,                                       &
-         sfmc=sfmc_lH, rzmc=rzmc_lH, tsurf=tsurf_lH, FT=FT_l, stemp=stemp_lH,    &
+         sfmc=sfmc_lH, rzmc=rzmc_lH, tsurf=tsurf_lH, asnow=asnow_l, FT=FT_l, stemp=stemp_lH,    &
```
gmao-rreichle, line 1501: I think this should read "asnow=asnow_lH" because "tile_data" is unbundled into the "_lH" arrays ("lH" is local with halo).  The same probably goes for "FT=FT_lH", which seems to be a bug in the current develop branch.  (Be careful relying on the "FT" implementation because it was never tested.)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -1497,6 +1511,7 @@ subroutine get_obs_pred(                                       &
     if (get_rzmc_lH)   rzmc_lH  = rzmc_l
     if (get_tsurf_lH)  tsurf_lH = tsurf_l
     if (get_FT_lH)     FT_lH    = FT_l
+    if (get_asnow_lH)  asnow_lH = asnow_l !jpark50
```
gmao-rreichle, line 1514: I'm honestly confused why we have this block of assignments.  Just above, we unpacked "tile_data_lH" into "*_lH", and now we seem to be overwriting "lH" with the simple local "l" array.  We'll have to take another close look at this later.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -3500,7 +3532,20 @@ subroutine cat_enkf_increments(                               &
...
+    real, dimension(N_catd, N_ens)        :: areasc_old 
+    real, parameter:: cpw   = 2065.22 !ice specific heat capacity @ 0C [J/kg/K]
+    real, parameter:: model_threshold = 0.40
+    real, parameter:: obs_threshold = 0.25
+    real, parameter:: wesn_added = 12.0 ! mm of SWE
+    real, parameter:: eps = 1.e-4
+    real, parameter:: small = 1.e-20
+    real, parameter:: rhofs = 150. ! kg/m^3 density of fresh snow
+    real, parameter:: rhoma = 500. ! kg/m^3 maximum snow density
+    real, parameter:: wemin = 13. ! kg/m^3
```
gmao-rreichle, line 3547: These parameters are used in the Stieglitz snow model and should be coming from there (rather than defined again here).  This should work with "use" statements as follows:

USE SurfParams,                 ONLY: WEMIN
USE StieglitzSnow,              ONLY: RHOMA, CPW
USE CATCH_CONSTANTS,   ONLY: RHOFS  => CATCH_SNWALB_RHOFS

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -3500,7 +3532,20 @@ subroutine cat_enkf_increments(                               &
     real, dimension(     N_catd)          :: tp1_ensavg
 
     type(obs_param_type)                  :: this_obs_param
-        
+   
+    !jpark50
+    real, dimension(N_catd, N_ens)        :: wesn_sum_old
+    real, dimension(N_catd, N_ens)        :: areasc_old 
+    real, parameter:: cpw   = 2065.22 !ice specific heat capacity @ 0C [J/kg/K]
```
gmao-rreichle, line 3539: Refer to the value in the StieglitzSnow module.  See comments about "use" statements below.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4000,7 +4045,120 @@ subroutine cat_enkf_increments(                               &
...
        ! ----------------------------------
-       
+    case (11) select_update_type ! 1d snow (asnow) analysis; MODIS snow cover fraction jpark50
+    !=============================================================================
+    !jpark50: 28-Jul-2021: For asnow update:
+    !The increment defined here is a direct replacement; there is no use of EnkF
+    !=============================================================================
+    if (logit) write (logunit, *) 'get 1d snow(asnow) increments; MODIS SCF'
+  
+    do n_e=1,N_ens
```
gmao-rreichle, line 4055: This call to "calc_asnow" appears to compute the forecast of the observation that are used below.  But this information should probably come from the "Obs_pred" (observation predictions), which have been calculated in "get_obs_pred()".  Note that "get_obs_pred()" was modified to compute "asnow" model forecasts, and "Obs_pred" is passed into this subroutine.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -5154,7 +5312,7 @@ end subroutine dist_km2deg
 !    
 !    implicit none
 !    
-!    integer, intent(in) :: N_ens, N_obs, N_catd
+!    integer, intent(in) :: N_ens, N_obs, N_catd
```
gmao-rreichle, line 5315: The new branch deletes a bunch of commented-out old code (Lines 5166-5221 in the develop branch).
But the clean-up isn't consistent.  I think we should leave this type of clean-up to a separate PR.  For now, it would be better to restore the deleted lines and focus on the snow DA changes.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -83,9 +83,439 @@ module clsm_ensupd_read_obs
...
+       
+       ! open and "start" hdf file 
+       file_id(j) = hopen( fnames(j), DFACC_READ, num_dds_block )  
+       status = vfstart(file_id(j))
+         
+       ! select vdata block that contains fields of interest
+       vdata_id(j) = vsfatch(file_id(j), vdata_ref, 'r') 
+       
+       ! determine number of records in vdata
+       status = vsqfnelt(vdata_id(j), n_records)
```
gmao-rreichle, line 188: I'm guessing the template for this subroutine was inherited from a reader for swath-based AMSR-E data.  In the case of the swath-based AMSR-E data, each file might have contained a different number of obs.  Here, however, each daily MODIS file contains a full global grid of 3600-by-7200 data points.  While querying "n_records" isn't wrong, it doesn't seem necessary.   That is, N_data_tmp could probably just be hard-wired as 3600*7200.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -83,9 +83,439 @@ module clsm_ensupd_read_obs
...
+ 
+    end if
+
+    allocate(lon(N_data))
+    allocate(lat(N_data))
+    allocate(MODIS_SCA(N_data))
+    
+    allocate(CI_Index(N_data))
+    allocate(Snow_QA(N_data))
+    allocate(Cloud_Index(N_data))
```
gmao-rreichle, line 213: Has N_data been initialized?  It's "intent(out)" and looks like it's uninitialized at this point.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -83,9 +83,439 @@ module clsm_ensupd_read_obs
...
+          lon(j)      = lon(i)
+          lat(j)      = lat(i)
+       end if
+       
+    end do
+   
+   if (logit) write (logunit, *) 'assimilation time', HH
+   if (logit) write (logunit, *) 'lat = ', lat
+   if (logit) write (logunit, *) 'lon = ', lon
+   if (logit) write (logunit, *) 'MODIS_SCA=', MODIS_SCA
```
gmao-rreichle, line 345: Shouldn't these write statements be limited to 1:j (where j is the final element of interest, see N_data=j in the next line)?  As they are, the write statements dump the entire allocated vectors into the file, incl whatever is left after the final N_data.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -83,9 +83,439 @@ module clsm_ensupd_read_obs
...
+   
+   integer, dimension(N_catd):: N_obs_in_tile
+
+   character(len=*), parameter :: Iam = 'read_obs_MODISsca'
+   character(len=400) :: err_msg
+
+   nullify (tmp_obs, tmp_lat, tmp_lon, tmp_tile_num)
+
+   write(HH, '(i2.2)') date_time%hour
+   read(HH, '(I10)') hh_num
```
gmao-rreichle, line 412: Why convert "date_time%hour" into a string and then back to an integer?  The modulus calculation below should work just fine as "mod(date_time%hour,3)".  Likewise for minutes and seconds, which should also be checked.

jpark50:
@gmao-rreichle @lcandre2 I pushed my local sandbox on Github for the purpose of update. I'm still keeping my eyes on the errors related to get_ind_obs() error (floating point exception). I will continue look details on tile_grid, Observation vector (e.g., lat, lon, tilenum).

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
gmao-rreichle: Comment probably outdated.  Closing for now.

gmao-rreichle:
@jpark50, @lcandre2:
I started going through the latest version of the branch.  For a start, I just pushed two commits that should not change anything.  The commits only clean up obsolete declarations, fix typos, fix indentation, and remove redundant code.  I will add a couple of in-line comments shortly and hope to take another, closer look tomorrow.

gmao-rreichle:
PS: I forgot to mention that it's best to hide whitespace changes when looking at the 2nd commit.  The indentation changes are otherwise very confusing.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -3408,8 +3438,12 @@ subroutine cat_enkf_increments(                               &
 
     ! -------------------------------------------------------------------
     
-    implicit none
+    USE SurfParams, ONLY: WEMIN
+    USE StieglitzSnow, ONLY: RHOMA_pub, cpw_pub
```
gmao-rreichle, line 3442: The GEOSldas branch (feature/jpark50/MODIS_snow_cover_fraction_assimilation) points to the "develop" branch of the GEOSgcm_GridComp repository, which does not have "RHOMA_pub" and "cpw_pub".  I suspect that @jpark50 made the necessary changes to StieglitzSnow in his sandbox.  This will ultimately need to be resolved if the variables are needed.  (It doesn't look like RHOMA_pub is used.)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4000,6 +4053,225 @@ subroutine cat_enkf_increments(                               &
...
+                      !Second options : replace the min function with if-else statement 
+                      if (deduction(n) .gt. 0.0) then
+                         TSN = 0.0
+                      else
+                         TSN = deduction(n)
+                      end if
+                      !if(logit) write(logunit,*) 'TSN = ',TSN
+                      !if(logit) write(logunit,*) 'wesn increment', cat_progn_incr(n,n_e)%wesn(i)
+                      !if(logit) write(logunit,*) 'cpw_pub: ', cpw_pub, 'MAPL_ALHF', MAPL_ALHF
+                      cat_progn_incr(n,n_e)%htsn(i) = real((TSN*cpw_pub-MAPL_ALHF))*cat_progn_incr(n,n_e)%wesn(i)
```
gmao-rreichle, line 4168: This looks to be terribly wrong, I'm afraid, and is a carry-over from Ally's implementation.  The idea of the snow analysis is to compute and apply increments for SWE first.  Then we need to adjust snow depth and snow heat content such that the snow density and snow temperature remain constant, resp.  If there wasn't snow prior to the analysis, we probably want to use what is appropriate for freshly fallen snow.  
If I'm reading the existing code correctly, the snow heat content is **always** based on the difference between the **air** temperature and the freezing point.  This is **not** what it should be.  
If there was some snow prior to the analysis, the snow heat content is not updated correctly.  To do this correctly, we need to compute the temperature of the snow before the analysis, and then find out the heat content so that the updated snow has the same temperature before and after the analysis.  This is not trivial because the ice fraction probably also plays a role.  
It's possible but unlikely that this causes the floating point that @jpark50 experienced.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4923,7 +5202,7 @@ subroutine check_compact_support(                                &
     
     select case (update_type)
 
-    case (1,3,4,5,6,9)   ! "1d" updates
+    case (1,3,4,5,6,9,11)   ! "1d" updates
```
gmao-rreichle, line 5205: The implementation of the snow cover analysis is based on the "1d filter" cases.  These cases worked in LDASsa up until the time when Clara used them in her MWR paper, which uses code from about 2017, well before GEOSldas came online.  It is possible that the old 1d filters no longer work in GEOSldas.  For GEOSldas development, we routinely tested and verified only the 3d Tb analysis of the SMAP L4_SM system.  That is, the floating point issue that @jpark50 is experiencing may be related to something that is not specific to the snow cover analysis.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4000,6 +4053,225 @@ subroutine cat_enkf_increments(                               &
...
+                         !if(logit) write(logunit,*) 'dens_layer', dens_layer
+                         
+                         cat_progn_incr(n,n_e)%sndz(i)=&
+                              cat_progn_incr(n,n_e)%wesn(i)/dens_layer
+                         
+                         !if(logit) write(logunit,*) 'cat_progn_incr(n,n_e)%sndz:', cat_progn_incr(n,n_e)%sndz(i)
+                         
+                      else
+                         dens_layer = RHOFS
+                         cat_progn_incr(n,n_e)%sndz(i) = cat_progn_incr(n,n_e)%wesn(i)/dens_layer
```
gmao-rreichle, line 4151: The equations that determine `cat_progn_incr(n,n_e)` here look suspicious.  Specifically, in the HISTORY output, "snow depth" is the depth of the snow within the snow covered fraction of the grid cell only, whereas "snow mass" is relative to the entire grid cell area (incl the snow-free and snow-covered portions).  See [MERRA-2 FAQ no. 2](https://gmao.gsfc.nasa.gov/reanalysis/MERRA-2/FAQ/#Q2).  I can't guarantee but I'm pretty sure that the same is true for the tile-space variables (%sndz, %wesn) here.  If so, then I would have expected "asnow" (forecast or analysis?) in the equations for cat_progn_incr.  Something to look at in more details later.  It's very unlikely that this has anything to do with Jongmin's run failure.

gmao-rreichle:
Continued in PR #512.  Closing without merge.

PR #437, opened 2021-08-11
fix to get_GEOSs2s to account for new PAR handling

gmao-jkolassa:
The proposed changes to get_GEOSs2s account for the new handling of PAR in the forcing data processing. The current default option is to compute PAR (and its components PARdrct and PARdffs) from SWdown **unless** the logical `PAR_available` is set to` .true. `within a forcing reader.

This PR introduces the variable `PAR_available` in get_GEOSs2s and sets it to `.true.` for FCST (for which PARdrct and PARdffs data exist in the forcing files) and keeps it as `.false.` for AODAS (for which PARdrct and PARdffs do not exits in the forcing files).

cc: @gmao-rreichle , @biljanaorescanin

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -2735,7 +2735,7 @@ contains
 
     if     (index(met_tag(8:11), 'FCST')/=0) then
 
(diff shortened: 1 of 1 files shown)


biljanaorescanin:
All tests passed. 
@gmao-rreichle, since @gmao-jkolassa already presented these results (bug fix solution)  and this is what she needs to make S2S reader work again should I go ahead and merge?

gmao-rreichle (approved):

PR #439, opened 2021-08-19
add unsetenv ld_library_path to  lenkf.j.template

saraqzhang:
unsetenv ld_library_path before source g5_modules to avoid having $fvroot/lib inherited from adas  in LADAS


gmao-rreichle:
@weiyuan-jiang,  please take a look at this change, which is needed when lenkf.j is used in the coupled land-atm data assimilation system.  Please approve if ok, or comment if not.  Thanks!

weiyuan-jiang (approved):

Issue #441, opened 2021-08-26
Fortran runtime error: Pointer argument 'ptrshforce' is not associated


mathomp4:
Last night I accidentally updated the LDAS nightly tests to use GCC 11.2. And in doing so, I seem to have triggered an exception:
```
Error termination. Backtrace:
At line 3693 of file /discover/nobackup/mathomp4/SystemTests/builds/LDAS_GNUCONUS/CURRENT/GEOSldas/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
Fortran runtime error: Pointer argument 'ptrshforce' is not associated
```
This is pointing to:
https://github.com/GEOS-ESM/GEOSldas/blob/61ada7dd54f75dc72d904d42df23367ee4defa8b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90#L3692-L3696

This is some pretty complex code, so I can't see anything obvious. I think you need to know the code path through this to know if this is a real bug in LDAS...or if we found a bug in GCC (always possible).

PR #442, opened 2021-08-26
get_GEOS() and get_GEOSs2s() fixes (corrected precip, pointer allocation)

weiyuan-jiang:
- Minor fix for GNU. Address issue #441 (check the pointer association before checking the size in get_GEOS() and get_GEOSs2s())
- Fix for allowing corrected precipitation to be on a different grid than rest of forcing data in get_GEOS().

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -2837,28 +2837,27 @@ contains
     
     do GEOSgcm_var = 1,N_GEOSgcm_vars
        
-       if (GEOSgcm_var==1) then
-          ! init shared memory
-          N_lon_tmp = -1
-          N_lat_tmp = -1
+       ! init shared memory
+       N_lon_tmp = -1
+       N_lat_tmp = -1
+       if (associated(ptrShForce)) then
+          N_lon_tmp = size(ptrShForce,1)
+          N_lat_tmp = size(ptrShForce,2)
+       endif
+       if(  (size(ptrShForce,1) /= GEOSgcm_grid_N_lon) .or.          &
+            (size(ptrShForce,2) /= GEOSgcm_grid_N_lat)       ) then
+          call MAPL_SyncSharedMemory(rc=status)
+          VERIFY_(status)
           if (associated(ptrShForce)) then
-             N_lon_tmp = size(ptrShForce,1)
-             N_lat_tmp = size(ptrShForce,2)
+             call MAPL_DeallocNodeArray(ptrShForce,rc=status)
+             VERIFY_(status)
           endif
-          if(  (size(ptrShForce,1) /= GEOSgcm_grid_N_lon) .or.          &
-               (size(ptrShForce,2) /= GEOSgcm_grid_N_lat)       ) then
-             call MAPL_SyncSharedMemory(rc=status)
-             VERIFY_(status)
-             if (associated(ptrShForce)) then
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
@weiyuan-jiang, 
I added two commits.

- The first commit supports the corrected precip being on a different grid from the rest of the forcing data in get_GEOS().   In a nutshell, I deleted the `if (GEOSgcm_var==1) then` condition per @gmao-qliu's email.  
- The second commit does the same for get_GEOSs2s(), which is not yet set up to read corrected precipitation from separate files, so here the change is not strictly needed, but it shouldn't hurt and will keep some consistency between the two subroutines.

In either case, did not change the order in which the forcing fields are processed (e.g., `G5DAS_defs`), contrary to my earlier plans.  On 2nd thought, I think re-ordering would not make a difference because the allocation is remembered from one time step to the next, so there is still only one back-and-forth change per forcing time step.  That is, re-ordering would only provide a minimal efficiency improvement in the initial or final time step.
I also renamed the PR and updated the introductory comment accordingly.  
Please  take a look when you're back in the office and let me know if you think this makes sense.  If everything looks ok, we can ask Biljana to test the PR.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -2837,22 +2837,27 @@ subroutine get_GEOSs2s(date_time, met_path, met_tag, N_catd, tile_coord, &
...
-       endif ! (GEOSgcm_var==1)
+          endif
+          call MAPL_AllocateShared(ptrShForce,(/GEOSgcm_grid_N_lon,GEOSgcm_grid_N_lat/),TransRoot= .true.,rc=status)
+          VERIFY_(status)
+          call MAPL_SyncSharedMemory(rc=status)
+          VERIFY_(status)
+       end if
+
        rc = 0
        call MAPL_SyncSharedMemory(rc=status)
```
gmao-rreichle, line 2862: Do we need these lines?  They seem redundant with the final two lines in the preceding if block.

weiyuan-jiang:
It looks good to me. @gmao-rreichle @biljanaorescanin

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
gmao-rreichle: Cleaned up with [`15c2763`](https://github.com/GEOS-ESM/GEOSldas/pull/442/commits/15c276386493b4304696f8af723d08467120e6a4)

gmao-rreichle:
Thanks, @weiyuan-jiang.  I committed two more changes, one to clean up the redundant call to MAPL_SyncSharedMemory() and the other to fix my earlier commit in get_GEOSs2s(). 
@biljanaorescanin : Please test the PR.  Make sure to use the full suite of tests that Matt is running every night.  One objective of the PR is to fix the pointer issue that Matt uncovered with the gnudebugconus test.

biljanaorescanin:
I've run same tests as Matt, all tests passed.

gmao-rreichle (approved):

gmao-rreichle:
@gmao-jkolassa, @jpark50:
FYI, with this PR I just merged a minor fix of the get_GEOSs2s() forcing reader (see intro comment above).  The same fix was applied and tested in get_GEOS(), so I don't expect any problems from this merge with get_GEOSs2s() going forward.  Please let me know if you encounter any problems when using this new version of get_GEOSs2s().  
cc: @weiyuan-jiang, @biljanaorescanin, @gmao-qliu

PR #448, opened 2021-09-14
Merge BRIDGE into main in prep for release v17.9.5

gmao-rreichle:


gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang, 
Can you please approve this PR but *not* merge it?  (I want to use a specific "squash and merge" message.)
- Rolf

biljanaorescanin (approved):

tclune (approved):
CMake changes ok.

mathomp4:
@gmao-rreichle One comment. You might want to change the `VERSION` number in the `CMakeLists.txt` file to correspond with the current version. Right now you have:
```cmake
project (
  GEOSldas
  VERSION 17.8
  LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
```
but that `VERSION` should probably be 17.9 at least.

gmao-rreichle:
Thanks, @biljanaorescanin and @tclune for approving the PR.    @mathomp4 : Thanks for the comment, which I only saw just after merging.   This has apparently been out of date for a while.  I'm not sure what the implications are of having the wrong version number in CMakeLists.txt.  I assume the wrong version number won't make a difference in any simulations.  Once I have Joe Ardizzone set up with the new release, I'll fix CMakeLists.txt in the "develop" branch so it's hopefully good for the next release.  
Thanks everyone for your speedy support in creating a new MAPL version that allows us to use bit shaving for SMAP L4_SM output. Much appreciated!

tclune:
For the moment there are no real consequences for having the wrong version in the CMakeLists.txt, and it is annoying that it must be maintained manually.

In theory, this version number can be used by "higher level" cmake projects so that they can specify the particular version of the sub project that should be used, or more often, the minimal number.   E.g., we might someday have logic in your fixture that does something like:
```cmake
find_package(MAPL 2.0.0<)
```
which  would indicate that MAPL must be at least at version 2.0.0.

As you might guess the greatest benefit will be low-level libraries where we may want to install multiple versions in the same directory tree and different projects can then select the version for their needs.    

I.e., nothing to be worried about, but nice to know these capabilities exist if we need them.

PR #449, opened 2021-09-14
Add auto tarball script

mathomp4:
This PR adds a new GitHub Action that on release, creates and uploads a tarball to GitHub. So if @gmao-rreichle  released v17.9.6, then in a minute or so, we should have a GEOSldas-v17.9.6.COMPLETE.tar.xz file that is completely mepo'd and everything.

Sadly, I missed the v17.9.5 release. Dang!

### DIFF
--- a/.github/workflows/release-tarball.yml
+++ b/.github/workflows/release-tarball.yml
@@ -1,0 +1,38 @@
+name: Create Complete Release Tarball
+
+on:
+  release:
+    types: [published]
+
+jobs:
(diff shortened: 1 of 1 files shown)


gmao-rreichle (approved):
Thanks, @mathomp4.  Sorry I just barged ahead with v17.9.5, propelled by the urgency of the SMAP production run.  It'll be great to have the auto tar ball in v17.9.6 and beyond.

PR #450, opened 2021-09-14
Update GEOSldas version info in CMakeLists.txt

gmao-rreichle:
@mathomp4, I hope this correction is what you had in mind for CMakeLists.txt.
Or would it make more sense to use  only the major version number ("v17") instead of "v17.9"?
Then we wouldn't have to manually update this info as often, although the lesser information might defeat the purpose.


gmao-rreichle:
Closing this PR.  I think I accidentally started my branch off "main", which gunks up the diff view.

Issue #398, opened 2021-05-10
Bilinear interpolation in ERA5_LIS forcing reader and lat/lon offset


gmao-rreichle:
As a future option, it may be worthwhile to **add bilinear interpolation** of the forcing data **to the ERA5_LIS forcing reader**.

This should also make the the following small lat/lon offset unnecessary:
https://github.com/GEOS-ESM/GEOSldas/blob/f00dbe3f85061f6e279b9e1296d4e255503f124e/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90#L1993-L2002
However, I now think the problem is tied to the nearest neighbor interpolation and is not specific ERA5.  That is, the unpredictable assignment of tiles to ERA5_LIS grid cells should also occur with GEOS-based forcing (which is also on a regular lat/lon grid) whenever nearest neighbor interpolation is used. 

No action is necessary right now.   We can revisit the bilinear interpolation for ERA5_LIS at a later time. 
cc: @weiyuan-jiang @biljanaorescanin

PR #463, opened 2021-09-27
Feature/wjiang/era5 h1

weiyuan-jiang:
@biljanaorescanin , This branch refactors your branch. Would you please help me test this branch against yours ? If they are zero-diff, please merge it.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -5406,124 +5406,26 @@ contains
     
     allocate(i1(N_cat),j1(N_cat))
     allocate(i2(N_cat),j2(N_cat),x1(N_cat),x2(N_cat),y1(N_cat),y2(N_cat))
+   
+    if (isCubed) then ! cs grid
+    ! cube-sphere grid of forcing data must match cube-sphere grid associated with tile space
+       i1(:) = tile_coord(:)%i_indg
+       j1(:) = tile_coord(:)%j_indg
+    else
+       i1(:) = ceiling((tile_coord(:)%com_lon - ll_lon)/dlon)
+       j1(:) = ceiling((tile_coord(k)%com_lat - ll_lat)/dlat)
+       ! NOTE: For a "date line on center" grid and (180-dlon/2) < lon < 180
+       !  we now have i1=(grid%N_lon+1)
+       ! This needs to be fixed as follows:
+       where( i1 > N_lon)  i1=1
+       if (any(j1 > N_lat)) then
+          err_msg = "encountered tile near the poles"
+          call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)
+       end if
+    endif
     
-    select case (m_hinterp)
+    call get_bilinear_neighbors(m_hinterp, tile_coord, i1, j1, ll_lon, ll_lat, dlon, dlat, N_lon, N_lat, i2, j2, x1, y1, x2, y2)
        
-    case (0)  ! nearest-neighbor
-       
-       ! compute indices for nearest neighbor interpolation from GEOSgcm grid to tile space
-       
-       if( isCubed ) then ! cs grid
-          
-          ! cube-sphere grid of forcing data must match cube-sphere grid associated with tile space
-          do k=1,N_cat
-             i1(k) = tile_coord(k)%i_indg
-             j1(k) = tile_coord(k)%j_indg
-          enddo
-
-       else
-
-          do k=1,N_cat
-             
-             ! ll_lon and ll_lat refer to lower left corner of grid cell
-             ! (as opposed to the grid point in the center of the grid cell)
-             
-             this_lon = tile_coord(k)%com_lon
-             this_lat = tile_coord(k)%com_lat
-             
-             i1(k) = ceiling((this_lon - ll_lon)/dlon)
-             j1(k) = ceiling((this_lat - ll_lat)/dlat)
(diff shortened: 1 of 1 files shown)


src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5442,34 +5362,64 @@ subroutine GEOS_openfile(FileOpenedHash, fname_full, fid, tile_coord, m_hinterp,
...
     !  i1, i2, j1, j2 (defining the indices of four neighbors)
+    isERA5_ = .false.
+    if (present(isERA5)) then
+       isERA5_ = isERA5
+    endif
+    offset  = 0.0
+    factor = 0.0
+    if (isERA5_) then
+       offset = 0.0001
+       factor = 0.5
```
gmao-rreichle, line 5398: I'm not sure this "factor" is correct/needed.  The ERA5 grid's lower left latitude (ll_lat) and longitude (ll_long) are already passed in correctly.  Maybe I'm missing something, but please make extra sure this is ok/needed.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5442,34 +5362,64 @@ subroutine GEOS_openfile(FileOpenedHash, fname_full, fid, tile_coord, m_hinterp,
...
     !  x1, x2, y1, y2 (defining the co-ords of four neighbors) and
     !  i1, i2, j1, j2 (defining the indices of four neighbors)
+    isERA5_ = .false.
+    if (present(isERA5)) then
+       isERA5_ = isERA5
+    endif
+    offset  = 0.0
+    factor = 0.0
+    if (isERA5_) then
+       offset = 0.0001
```
gmao-rreichle, line 5397: In the description of the issue #398, I was wondering if the "offset" is indeed specific to ERA5.  It may also needed for GEOS forcing data.  Having the "offset" should only make a noticeably difference for nearest-neighbor interpolation, which we never use for GEOS forcing.   Making things consistent, however, might result in a non-zero-diff change.  This probably requires a closer look and more discussion.

gmao-rreichle:
Thanks for putting this together, @weiyuan-jiang.  I added two in-line comments above.  The potential for ending up with a non-zero-diff change might mean that we have to test in steps.  Perhaps the first thing to do is for @biljanaorescanin to do two zero-diff tests against her branch using ERA5_LIS forcing, one for nearest neighbor and one for bilinear interpolation.  This should resolve my question about "factor".  Once we know the changes are indeed zero-diff  for ERA5_LIS, we can do the usual nightly tests vs. develop.  Thereafter, we should look into "offset" and see if this needs changing and may result in a desirable (clean) change that is non-zero-diff vs develop.

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
weiyuan-jiang: @biljanaorescanin Which group is using  ERA5 grid ? where do you get this extra 0.5*dlon (dlat)?

biljanaorescanin:
I've tested: 
My branch EAR5 option with nearest neighbor interpolation vs this one its zero diff for "lfs"  and "lnd" collections. 
Same goes for test with bilinear interpolation, its zero diff.

biljanaorescanin:
Nightly tests are zero diff.

weiyuan-jiang:
@biljanaorescanin , I have pushed the fix that we talked about during the meeting. Would you please test it again?

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5540,6 +5362,84 @@ subroutine GEOS_openfile(FileOpenedHash, fname_full, fid, tile_coord, m_hinterp,
...
+    endif
+
+    x1 = (i1-1)*dlon + ll_lon  + dlon/2.
+    y1 = (j1-1)*dlat + ll_lat  + dlat/2.
+    
+    tmp_lon = tile_coord%com_lon + dlon/2.
+    tmp_lat = tile_coord%com_lat + dlat/2.
+   
+    inew =  ceiling((tmp_lon  - ll_lon)/dlon)   ! nearest-neighbor i index of shifted location
+    jnew =  ceiling((tmp_lat  - ll_lat)/dlat)   ! nearest-neighbor j index of shifted location
```
gmao-rreichle, line 5417: I can't say for sure, but I think to be consistent we also need to apply "offset" in this calculation.
@biljanaorescanin's branch for ERA5_LIS bilinear interpolation also doesn't  do this consistently.
Perhaps that's why experiments with M-2 forcing have roundoff diffs vs. this branch.  (Note that "develop" uses the same "this_lon" and "this_lat" in both "ceiling()" operations for determining (i) the nearest neighbor and (ii) the grid cell that sits diagonally across from the nearest neighbor).
Please add "+offset" into lines 5416 and 5417 and try the M2 experiment again. 

If this doesn't work, we'll need to keep looking.  I note that the implementation of get_neighbor_index() is somewhat different from the corresponding code in the original GEOS_OpenFile() on develop -- which is what we're comparing against when running with M2 forcing.  There may still be a difference in the placement of the "wrap-around" statements between get_neighbor_index() and GEOS_OpenFile() on develop.

biljanaorescanin:
This change had no impact to the results.

PR #464, opened 2021-10-01
Updates for Cascade Lakes at NCCS

mathomp4:
This is an update to GEOSldas that is needed to use Cascade Lake nodes at NCCS. (There is a similar PR in for GEOSgcm, see https://github.com/GEOS-ESM/GEOSgcm/pull/335)

First, ESMA_env is moved to v3.4.0 which moves the Intel compiler to Intel 2021.2 (needed on Cascade Lake). Next, ESMA_cmake v3.6.0 updates the Intel `Release` flags to be vectorized everywhere. Testing with GEOSgcm has shown that these changes are indeed zero-diff (and faster than old `Release` flags). However, I have not yet tested this with the GEOSldas. @weiyuan-jiang or @biljanaorescanin might be best equipped to do this.

I've also taken a stab at updating `ldas_setup` for the Cascade Lakes. I probably got it...close? I suppose @gmao-rreichle and @weiyuan-jiang can take a look. 

I'll keep this a draft until all necessary people can chime in. I suppose the two main questions are:

1. Is the update to Intel 2021.2 and the change to `Release` flags zero-diff?
2. Is `ldas_setup` correctly set up to work on Cascade Lake?
3. Is LDAS any faster in `Release` mode with the new flags?

As such, I've labeled this as 0-diff and not 0-diff and put a blocker on it.

### DIFF
--- a/src/Applications/LDAS_App/ldas_setup
+++ b/src/Applications/LDAS_App/ldas_setup
@@ -1510,7 +1510,7 @@ def _printRmInputKeys(rqdRmInpKeys, optRmInpKeys):
     print '# - walltime         = walltime requested; format is HH:MM:SS (hours/minutes/seconds)'
     print '# - ntasks_model     = number of processors requested for the model (typically 112; output server is not included)'
     print '# - ntasks-per-node  = number of tasks per node (typically 48 for cascade lake, 40 for skylake, and 28 for haswell nodes)'
-    print '#                      [If >48, cascade lake nodes will be allocated, if >40, cascade or skylake, else cascade, skylake or haswell.]'
+    print '#                      [If >40, cascade lake nodes will be allocated, if >28, cascade or skylake, else cascade, skylake or haswell.]'
     print '#'
     for key in rqdRmInpKeys:
         print key + ':'
--- a/src/Applications/LDAS_App/ldas_setup
+++ b/src/Applications/LDAS_App/ldas_setup
@@ -1241,8 +1241,8 @@ class LDASsetup:
                     elif 'MY_NODES' in line :
                         line_ = line.replace('MY_NODES',str(self.optRmInp['nodes']))
                         fout.write(line_.replace('MY_NTASKS_PER_NODE',str(self.rqdRmInp['ntasks-per-node'])))
-                        if int(self.rqdRmInp['ntasks-per-node']) > 28:
-                           fout.write("#SBATCH --constraint=sky\n")
+                        if int(self.rqdRmInp['ntasks-per-node']) > 40:
+                           fout.write("#SBATCH --constraint=cas\n")
                     elif 'MY_OSERVER_NODES' in line :
                         fout.write(line.replace('MY_OSERVER_NODES',str(self.optRmInp['oserver_nodes'])))
                     elif 'MY_WRITERS_NPES' in line :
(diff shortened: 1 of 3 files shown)


gmao-rreichle:
Thanks for putting this together, @mathomp4.
@weiyuan-jiang: Matt included the ldas_setup changes for Cascade Lake nodes that you were going to add per our Wed meeting. Matt's branch looks good to me, but please double-check it is consistent with what you would have done. 
@biljanaorescanin: When you get a chance, please run the full set of tests for this PR.
I'm keeping my fingers crossed that it will be 0-diff, but I wouldn't be surprised if the LDAS turns out to be more complicated than the GCM in this context.

src/Applications/LDAS_App/ldas_setup
```diff
@@ -1241,8 +1241,8 @@ class LDASsetup:
                     elif 'MY_NODES' in line :
                         line_ = line.replace('MY_NODES',str(self.optRmInp['nodes']))
                         fout.write(line_.replace('MY_NTASKS_PER_NODE',str(self.rqdRmInp['ntasks-per-node'])))
-                        if int(self.rqdRmInp['ntasks-per-node']) > 28:
-                           fout.write("#SBATCH --constraint=sky\n")
+                        if int(self.rqdRmInp['ntasks-per-node']) > 40:
+                           fout.write("#SBATCH --constraint=cas\n")
```
weiyuan-jiang, line 1245: if tasks-per-node is between 28 and 40,  the constrain should be sky. Without constrain, it goes to haswell and may cause error.

weiyuan-jiang:
```suggestion
                      if (28 < int(self.rqdRmInp['ntasks-per-node']) and int(self.rqdRmInp['ntasks-per-node']) <= 40) :
                         fout.write("#SBATCH --constraint=sky\n")
```

weiyuan-jiang:
```suggestion
                           fout.write("#SBATCH --constraint=cas\n")
                        if (28 < int(self.rqdRmInp['ntasks-per-node']) and int(self.rqdRmInp['ntasks-per-node']) <= 40) :
                           fout.write("#SBATCH --constraint=sky\n")```

biljanaorescanin:
After @mathomp4 last update all tests are failing, so this this will not work. But, before we passed 4 tests until new change was pushed. Since I started running tests as soon as I've seen PR. I will reject this last change and try to run tests with previous commit.

gmao-rreichle:
@biljanaorescanin : If I understand your comment correctly, the tests are failing during ldas_setup.  The more important tests are to see if the Intel compiler upgrade and the new release flags are 0-diff.   Because this involves only the Intel compiler (I think), I suggest you try Matt's branch but with ldas_setup from develop and only for the Intel tests.

mathomp4:
Note: I can never guarantee that **_Aggressive_** results will ever be zero-diff upon changing compilers. They are aggressive flags and Intel often optimizes/changes what their fancy flags can do. I just wanted speed for speed's sake so, I never stare too hard if things change. If the world still looks like Earth, I figure it's okay.

gmao-rreichle:
@mathomp4, @weiyuan-jiang, @biljanaorescanin:
Let me try to recap where things stand with this PR.

i) The ldas_setup problem from Biljana's first round of tests was resolved by Weiyuan.
ii) Biljana's most recent, Intel-only tests show that we do indeed get zero-diff for the Intel compiler as long as we're not using Aggressive optimization.  The latter yields roundoff-type differences, as expected.
iii) Per Weiyuan's email, the nightly test didn't compile with GNU, even though the fancy new Release flags are for Intel only (or so at least I think).  I'm not sure what the problem might be. 

Based on i) and ii), I'd be ready to move forward.  But what about iii)?   What am I missing?

mathomp4:
@gmao-rreichle I'm not sure why the GNU tests failed. They shouldn't have. No changes happened to GNU as far as I can tell. My guess is that maybe Weiyuan needs to update the GNU modules being used in his testing setup. I'm using:

  6.2.7/g5_modules.gcc11

and his might be older than that?

biljanaorescanin:
@weiyuan-jiang did tests on Cascade and I run them on Sky we both got zero diff for all tests except agressive which we expected to see.

PR #465, opened 2021-10-04
Make bit shaving default in HISTORY template 

gmao-rreichle:
Make bit shaving default in HISTORY template (except for "catch_progn_incr" Collection).

TBD: nbits=10 or nbits=12
FP typically uses nbits=10, MERRA-2 used nbits=12.  Waiting to see if Rob L can find something about this in his records. 

PR is **not** 0-diff for HISTORY output but should be 0-diff for restarts.  

UPDATED 6 Oct 2021:
- nbits is the number of bits **retained**.  This PR makes the GEOSldas default consistent with the less aggressive shaving of MERRA-2 (nbits=12).
- Clarifying comment added to GEOSldas_HIST.rc.
- 0-diff depends on whether Matt's nightly test scripts use GEOSldas_HIST.rc from the src code or supply separate copies of HISTORY.rc files the tests.   For now, assuming the latter and switching flag from non-0-diff to 0-diff.

@biljanaorescanin:  When you get a chance, please run the standard Intel tests only (conus, global, globalcs, debugconus).  No need to run the "aggressive" tests or the GNU tests.  If the tests pass, please approve and merge the PR.

FYI:  Matt found the following [poster on bit shaving](https://www.essoar.org/doi/pdf/10.1002/essoar.bb4d390d7f9f73ba.963021f17c584ca8.1)

UPDATED 13 Oct 2021:
- @biljanaorescanin tested and found **0-diff for restarts but non-0-diff for HISTORY (lnd and lfs) output**, consistent with the use of this template in Matt's nightly tests, which need to be re-baselined after this PR is merged.  
- EXPID becomes part of the file name
- EXPDSC and EXPSRC are expected by MAPL but not currently used
- clarified comments about EXPID, EXPDSC, and EXPSRC, see [MAPL issue 1080](https://github.com/GEOS-ESM/MAPL/issues/1080)


biljanaorescanin:
I can confirm for all tests restarts are zero diff. Only differences are seen in "lfs" and 'lnd" collections ( for all tests) which we expected to see due to the bit shaving.

Issue #480, opened 2021-10-27
Bug in IMS.rc for LatLon tile space


gmao-rreichle:
The `IMS.rc` file defines the decomposition of the simulation domain into vertical stripes for tile spaces based on EASE grids and LatLon grids.  (For tile spaces based on cube-sphere grids, `JMS.rc` is used.)

The comments in [subroutine optimize_latlon() in preprocess_ldas_routines.F90](https://github.com/GEOS-ESM/GEOSldas/blob/dc835bc9c29e7cb5fff4c405c89470c91d690626/src/Applications/LDAS_App/preprocess_ldas_routines.F90#L2020) state that `IMS(k)>=2`, but this condition is not ensured during pre-processing.  (For cube-sphere tiles, `JMS(k)>=2` is ensured.)

**A bug fix in [subroutine optimize_latlon()](https://github.com/GEOS-ESM/GEOSldas/blob/dc835bc9c29e7cb5fff4c405c89470c91d690626/src/Applications/LDAS_App/preprocess_ldas_routines.F90#L2020) is needed so that `IMS(k)>=2` for EASE and LatLon tile spaces.**  Also, if `IMS(k)>=2` **_or_** `JMS(k)>=2` cannot be ensured for **all** `k` (so that there will be no "empty" processors), the pre-processor should exit with a message telling the user to request fewer processors.

_**Background:**_

The `[I,J]MS>=2` condition is needed by MAPL to generate **2d** (gridded) HISTORY output.   

For now, `IMS(k)=1` would be ok for tile spaces based on the EASE grid because only 1d (tile-space) output is available and 2d output is not (yet) available for these tile spaces, [see README.OutputSpecs.md](https://github.com/GEOS-ESM/GEOSldas/blob/develop/doc/README.OutputSpecs.md).  

However, `IMS(k)=1` occurs only when (too) many processors are used, especially over small domains.  For the SMAP L4_SM use case with 120 processors, IMS(k)>=12.  For the CONUS test case with 28 processors, IMS>=4.  Indications are that using more processors is not efficient and should be discouraged.

Moreover, testing by @weiyuan-jiang suggests that when `IMS(k)>=2`, _binary_ 2d output can be generated, although this output has not been validated, and more work would be needed to ensure that conversion to nc4 works using [tile_bin2nc4.F90](https://github.com/GEOS-ESM/GEOSldas/blob/develop/src/Applications/LDAS_App/tile_bin2nc4.F90).

[...]

PR #481, opened 2021-11-01
enforce IMS >=2

weiyuan-jiang:
This PR is addressing issue #480

### DIFF
--- a/src/Applications/LDAS_App/preprocess_ldas_routines.F90
+++ b/src/Applications/LDAS_App/preprocess_ldas_routines.F90
@@ -2220,12 +2220,12 @@ contains
           enddo
           local = j
        enddo
-      if( sum(JMS) /= JMGLOB) then
+       if( sum(JMS) /= JMGLOB) then
           print*, sum(JMS), JMGLOB
           stop ("wrong cs-domain distribution in the first place")
        endif 
-       ! adjust JMS.rc make make sure no processor ("stripe") has 0 or 1 grid cells in j dimension
-       ! (i.e., each proc's subdomain must include at least 2 grid cells in the j dimension;
+       ! adjust JMS.rc to make sure each processor has at least 2 grid cells in j dimension 
+       ! (i.e., each proc's subdomain must include at least 2 latitude stripes;
        !  stripes of grid cells may or may not contain land tiles)
        j = 1
        do k = 1,6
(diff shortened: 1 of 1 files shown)


src/Applications/LDAS_App/preprocess_ldas_routines.F90
```diff
@@ -2438,7 +2438,7 @@ subroutine optimize_latlon(fname_tilefile, N_proc_string)
        
        if( sum(local_land) /= total_land) stop ("wrong distribution")
        if( sum(IMS) /= N_lon) stop ("wrong domain distribution")
-       
+       if( any(IMS <=1) ) stop ("Ask for too many processors. Each processor should have at least 2 grid cells")
```
gmao-rreichle, line 2441: @weiyuan-jiang, 
Thanks for taking a crack at this issue.  This fix isn't quite what I had thought was needed.  
Let's say we end up with: 
`IMS = [ 2 2 2 2 1 3 2 2 ]`
for 16 longitude strips on 8 processors, then a simple solution would be to re-assign one longitude strip so we have 
`IMS = [ 2 2 2 2 2 2 2 2 ]`
This is perhaps slightly less optimal than if we were able to allow `IMS(5)=1`, but it should be serviceable.  As implemented now, the user's only option would be to use, say, 4 processors and probably end up with something like:
`IMS = [ 4 4 5 3]`
I think for JMS on the cube-sphere tile space, we already have a fix for JMS(k)=0 or JMS(k)=1 (~line 2235):
```
if(JMS(i) == 0) then
   n = maxloc(JMS(n1:n2),DIM=1)
   JMS(i) = 1
   JMS(n+n1-1) = JMS(n+n1-1)-1
endif
if(JMS(i) == 1) then
   n = maxloc(JMS(n1:n2),DIM=1)
   JMS(i) = 2
   JMS(n+n1-1) = JMS(n+n1-1)-1
endif
```

weiyuan-jiang:
Ok.  Now it redistributes IMS before it exits.

gmao-rreichle:
@biljanaorescanin, when you get a chance, please test this PR.  The changes are limited to the IMS.rc pre-processing and should be 0-diff (at least with standard optimization), but it's best to test just in case.

biljanaorescanin:
All tests passed.

Runtype         Clone Build Build Time Model Run/Compare Assim Run/Compare
-------         ----- ----- ---------- ----------------- -----------------
conus           pass  pass  12 min     pass/pass          -- / --        
global           --    --    --        pass/pass         pass/pass        
globalcs         --    --    --        pass/pass         pass/pass        
globalcnclm4     --    --    --        pass/pass          -- / --        
debugconus       --   pass   4 min     pass/pass          -- / --        
aggconus         --   pass  12 min     pass/pass          -- / --        
aggglobal        --    --    --        pass/pass         pass/pass        
aggglobalcs      --    --    --        pass/pass         pass/pass        
aggglobalcnclm4  --    --    --        pass/pass          -- / --        
gnuconus        pass  pass  11 min     pass/pass          -- / --        
gnuglobal        --    --    --        pass/pass         pass/pass        
gnuglobalcs      --    --    --        pass/pass         pass/pass        
gnuglobalcnclm4  --    --    --        pass/pass          -- / --        
gnudebugconus    --   pass   8 min     pass/pass          -- / --

gmao-rreichle (approved):

PR #484, opened 2021-11-05
MAPL V3 FLAP Simplification

Gvilla1000-nasa:
This PR adds a necessary change for LDAS so it works with MAPL-v3 when that is released.


Gvilla1000-nasa:
@mathomp4 tagging you for your awareness.

mathomp4 (approved):
Approving for MAPL3

PR #488, opened 2021-12-16
change MAPL version to fix nbits

weiyuan-jiang:
Change env, cmake and MAPL version to reflect the fix for bit shaving ( normal return when a processor has no tile data)


weiyuan-jiang:
will be addressed by another PR.

PR #487, opened 2021-12-16
Update to MAPL v2.14.0, ESMA_env v3.7.0, and ESMA_cmake v3.7.3;  fix timers

weiyuan-jiang:
The timers are fixed so the "on" and "off" can match

PR #489, opened 2021-12-16
Use MAPL patch to fix nbits

biljanaorescanin:
Running develop version of GEOSldas gave error in GEOSldas_err_txt: 
"ShaveMantissa32: Bad length of mask bits: len= 0, xbits= 12" 
@weiyuan-jiang found there was issue in MAPL when we use too many processors if not enough tiles. 
Within MAPL 2.10.1 patch he added fix to return 0 when there is no data for bit shaving. 
To make new MAPL work we had to move to newer cmake v3.6.5 and env v3.4.1

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -5,13 +5,13 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v3.4.0
+  tag: v3.4.1
   develop: main
 
 cmake:
   local: ./@cmake
   remote: ../ESMA_cmake.git
-  tag: v3.6.0
+  tag: v3.6.5
   develop: develop
 
 ecbuild:
@@ -29,7 +29,7 @@ GMAO_Shared:
 MAPL:
   local: ./src/Shared/@MAPL
   remote: ../MAPL.git
-  tag: v2.8.6
+  tag: v2.10.1
   develop: develop
 
 GEOSgcm_GridComp:


biljanaorescanin:
All Nightly tests passed. 
GEOSldas run that used to give nbits error now works.

mathomp4:
> All Nightly tests passed. GEOSldas run that used to give nbits error now works.

And soon...another PR just like this with new version numbers! Hopefully it does the same! :smile:

gmao-rreichle (approved):

gmao-rreichle:
Thanks, @biljanaorescanin, @weiyuan-jiang and @mathomp4.
I approved and merged this PR as a stepping stone.  Since we already have MAPL 2.14.0 and an associated PR #487, let's test #487.  If #487 is 0-diff, we can merge and create the new GEOSldas tag based on MAPL 2.14.0.  I'll add comments in #487.

mathomp4:
> Thanks, @biljanaorescanin, @weiyuan-jiang and @mathomp4. I approved and merged this PR as a stepping stone. Since we already have MAPL 2.14.0 and an associated PR #487, let's test #487. If #487 is 0-diff, we can merge and create the new GEOSldas tag based on MAPL 2.14.0. I'll add comments in #487.

@gmao-rreichle Sounds good. Note that the next time we update env and cmake it'll be non-zero-diff but it's not because MAPL needed it. Rather it's newer compilers and changing some flags. 

But for now at least, we have good enough compilers that I could keep the older compiler-flag combo going for a while until you have time to validate.

PR #492, opened 2021-12-20
fix timers, update components, update CHANGELOG.md

weiyuan-jiang:
fix timers and update components. Replaces #491 and #487.


gmao-rreichle:
successfully tested for 0-diff by @weiyuan-jiang

gmao-rreichle (approved):

PR #505, opened 2022-01-03
remove catchincr grid comp when it is not land_assim

weiyuan-jiang:
UPDATED by @gmao-rreichle:
This PR fixes a problem reported by @gmao-qliu on 3 Jan 2022:
When GEOSldas v17.10.0 is used to write out L-band Tb data in a single-member simulation without perturbations and without data assimilation (ie, land_assim=no), the run crashes during MAPL_GenericInitialize() in the LandAssim GridComp (when the mwRTM parameters are read from the restart file).  There is no crash when v17.10.0 is used in an ensemble data assimilation configuration. 

The single-member, no assimilation configuration with Tb output is needed for the SMAP Nature Run (e.g., NR5000) that is delivered to the SMAP Project at JPL.  The last tag for which this is known to work is v17.9.0-beta.8 (currently used for NR5000).  

---
@saraqzhang Would you please verify this PR for me? The question here is if there exists such scenario that it is not land_assim but needs catchincr grid comp?

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -982,29 +982,31 @@ contains
       RC=STATUS)
 
 !
-   call MAPL_GetResource ( MAPL, NUM_ENSEMBLE, Label="NUM_LDAS_ENSEMBLE:", DEFAULT=1, RC=STATUS)
-   _VERIFY(STATUS)
-   call MAPL_GetResource ( MAPL, FIRST_ENS_ID, Label="FIRST_ENS_ID:", DEFAULT=0, RC=STATUS)
-   _VERIFY(STATUS)
-   call MAPL_GetResource ( MAPL, ens_id_width, Label="ENS_ID_WIDTH:",      DEFAULT=0,       RC=STATUS)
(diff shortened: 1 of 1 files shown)


saraqzhang:
@weiyuan-jiang    If not land_assim , no analysis increments will be produced.  I don't think catchincr gc is needed  in that no assim mode.

gmao-rreichle:
@weiyuan-jiang, 
I agree with @saraqzhang that we shouldn't need the CatchIncr GridComp if land_assim==false.
To be safe, though, I think @biljanaorescanin should test the PR. 
Also, is this PR related to the error that @gmao-qliu reported to me today (ie, MAPL_GenericInitialize() for the Landassim GC fails if land_assim==false, in case we just want to export Tb but not do Tb data assimilation)?

gmao-qliu:
> @weiyuan-jiang, I agree with @saraqzhang that we shouldn't need the CatchIncr GridComp if land_assim==false. To be safe, though, I think @biljanaorescanin should test the PR. Also, is this PR related to the error that @gmao-qliu reported to me today (ie, MAPL_GenericInitialize() for the Landassim GC fails if land_assim==false, in case we just want to export Tb but not do Tb data assimilation)?

gmao-qliu:
I believe this is related to the issue that I reported today.

biljanaorescanin:
All tests have passed.

gmao-rreichle (approved):

PR #512, opened 2022-01-10
MODIS snow cover fraction assimilation

gmao-rreichle:
Jul 10, 2023 update

Clean up and modifications to MODIS SCF assimilation:
1.  Modified the MODIS_SCF  in LDASsa_DEFAULT_inputs_ensupd.nml to reflect recent changes to data location and switch from SCA to SCF to be more accurate.
2. Included HTSNN1-3 and Albedo parameters in tile_bin2nc4.F90 to include necessary information in nc4 outputs needed for assimilation verification. 
3. Reorganized and cleaned up clsm_ensupd_read_obs.F90, including improved variable names and organization. Maintained the 1D restructuring of MODIS data.
4.  Modified clsm_ensup_upd_routines.F90 by removing most hard coded terms and renaming variables to be more specific to snow cover assimilation. There are 3 outstanding issues:
             * The apha, beta, and max_incr_swe terms are still hard coded within upd_routines. Rolf, Qing and I discussed where these should ultimately go, ideally in a location that can be changed at run time. Rolf will look into this.
             * Hard coded snow layer target thickness partially remains (Line 3519) The max top layer thickness can be imported from catch_constants.F90; however, the remaining values, which should be the distribution of thickness in layers 2 through N_snow in terms of fraction. These could be calculated within upd_routines as 1.0/real(N_snow-1, kind=4) following Stieglitz or could alternatively be exported along with the top layer thickness. In my mind, this decision depends on what relayer2 really needs. Does it simply need reasonable values (0.08 0.5 0.5) or should it be using the forecast values? 
             * When there is snow in the forecast and the analysis, the code currently calculates the swe and heat content of each layer based on the swe ratio; however, there could be instances where the forecast snow layer is at 0 deg C and has liquid water within it. So, it may be more correct to calculate the heat content as a function of the forecast snow temperature and frozen fraction, which would look something like what I include in a comment on line 4568. I'd be interested in anyone's thoughts.  
5. Slight modifications to clsm_ensupd_enkf_update.F90 for clarity.

tagging @gmao-rreichle, @saraqzhang, @gmao-qliu, @amfox37 

%%%%%%%
This PR continues where #434 left off.

Starting from Jongmin's branch (`feature/jpark50/MODIS_snow_cover_fraction_assimilation`), I removed three bad and redundant assignments in the MODIS SCA obs reader.  
I also created an associated GEOSgcm_GridComp branch and edited the components.yaml file accordingly.  The GEOSgcm_GridComp branch and its connection to the GEOSldas branch is not meant to be final and will need to be revisited. 
These changes were implemented in Jongmin's sandbox but hadn't made it into his branch.
 
For reference, I copied Jongmin's sandbox and experiments to here:
/discover/nobackup/rreichle/Jongmin_SCA_DA/from_Jongmin/

cc: @lcandre2, @weiyuan-jiang, @gmao-qliu


tclune (dismissed):
cmake changes ok

weiyuan-jiang:
@jpark50 I just fixed your read_obs according to my understanding of your program ( I may be wrong though, please double check).  But I believe the more serious error is the dimension of mat_force that you passed into the subroutine cat_enkf_increments.   The number N_catd and nTiles_ana are different if I understand correctly. You need to have a forcing that matches nTiles_ana
@gmao-rreichle

gmao-rreichle:
I just committed what I hope is a fix for the "met_force" dimension error: [e8083f2](https://github.com/GEOS-ESM/GEOSldas/pull/512/commits/e8083f2ec3c730bad7415ad733d791a1d3ac3e61).  In a nutshell, "met_force" matches the decomposition of the model, but in the call to cat_enkf_increments(), the decomposition is that after the "Analysis Load Balance". 
@gmao-qliu, @weiyuan-jiang: Can you please coordinate so one of you can try out this fix? Thanks! 
cc: @lcandre2

weiyuan-jiang:
Now the program passes that point and crashes here:
forrtl: error (75): floating point exception
Image              PC                Routine            Line        Source             
GEOSldas.x         0000000000766F9B  Unknown               Unknown  Unknown
libpthread-2.22.s  00002B1FD61E9CE0  Unknown               Unknown  Unknown
libMAPL.base.so    00002B1FCC84290F  Unknown               Unknown  Unknown
GEOSldas.x         00000000006EC6D7  sfclayer_mp_helfs         380  surfacelayer.F90
GEOSldas.x         000000000061633E  geos_catchgridcom        3436  GEOS_CatchGridComp.F90

weiyuan-jiang:
Actually, this error comes earlier. It is buried in a long print out info.
forrtl: error (63): output conversion error, unit 26, file /gpfsm/dnb32/wjiang/LDAS_exp/model/SCA_ASSIM_reg/scratch/SCA_ASSIM_reg.SMAP_L4_SM_gph.ctl
Image              PC                Routine            Line        Source             
GEOSldas.x         0000000001178F0B  Unknown               Unknown  Unknown
GEOSldas.x         00000000011D9ED7  Unknown               Unknown  Unknown
libMAPL.history.s  00002B17A38086EB  mapl_historygridc        3991  MAPL_HistoryGridComp.F90
libMAPL.history.s  00002B17A37F46A0  mapl_historygridc        3615  MAPL_HistoryGridComp.F90
libesmf.so         00002B17AB0B42E4  _ZN5ESMCI6FTable1     Unknown  Unknown
libesmf.so         00002B17AB0B7ABF  ESMCI_FTableCallE     Unknown  Unknown
libesmf.so         00002B17AB644D45  _ZN5ESMCI2VM5ente     Unknown  Unknown
libesmf.so         00002B17AB0B573A  c_esmc_ftablecall     Unknown  Unknown
libesmf.so         00002B17AB8D4D4D  esmf_compmod_mp_e     Unknown  Unknown
libesmf.so         00002B17ABBD50C1  esmf_gridcompmod_     Unknown  Unknown
libMAPL.base.so    00002B17A4558868  mapl_genericmod_m        1835  MAPL_Generic.F90

gmao-rreichle:
> Now the program passes that point and crashes here: forrtl: error (75): floating point exception ...

Thanks, @weiyuan-jiang.  Based on the crash being in the surface layer turbulence alone, my first guess is that a snow variable was updated into something unphysical, and that the updated snow state still needs checking to avoid unphysical states.  It'd be helpful to know exactly which variable results in the floating point exception, and what it's value is.

gmao-rreichle:
> Actually, this error comes earlier. It is buried in a long print out info. ( It is not important at this moment)
> forrtl: error (63): output conversion error, unit 26, file  /gpfsm/dnb32/wjiang/LDAS_exp/model/SCA_ASSIM_reg/scratch/SCA_ASSIM_reg.SMAP_L4_SM_gph.ctl

@weiyuan-jiang: I cannot find this log entry anywhere.  Also, it's about a GRADS control file that is only used only for post-processing.  In any case, the file exists, but it has a problem in the following line:

`xdef ***** linear    0.000     1.000000000`

The ***** should be an integer with the number of tiles.  Looks like a bug in Jongmin's code, or perhaps just in his MAPL HISTORY.rc file.  I'll keep looking. 
 
Having said that, the error in `./scratch/GEOSldas_err_txt` about the floating point exception in `surfacelayer.F90` is probably more relevant.

weiyuan-jiang:
Some values of psl are smaller than 0 , which is not allowed to pass into next call here
https://github.com/GEOS-ESM/GEOSgcm_GridComp/blob/fc71edb3500bc960a4d3ba3592eba265606c1090/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/GEOScatch_GridComp/GEOS_CatchGridComp.F90#L3433-L3434

lcandre2:
I worked on updating the updates to wesn, sndz, and htsn in clsm_ensupd_upd_routines.F90. There are still substantial issues but I would like to discuss lines 4518:4619.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4430,13 +4488,177 @@ subroutine cat_enkf_increments(                               &
...
+                   ! ADD SNOW: SCA of model is less than observed SCA
+                   
+                   swe_incr(kk,n_e)                                                                     &
+                        = max_incr_swe * (Observations(ind_obs(1))%obs - asnow_fcst/alpha_threshold)
+                   
+                elseif (Observations(ind_obs(1))%obs .lt. beta_threshold) then
+                   
+                   ! REMOVE SNOW: IF SCA of model is greater than observed SCA
+                   
+                   swe_incr(kk,n_e) = (-1.) *  max_incr_swe * (1. - Observations(ind_obs(1))%obs/beta_threshold)
```
gmao-rreichle, line 4552: After thinking about this overnight I realized that the "remove snow" equation should read as follows:
`swe_incr(kk,n_e) = (-1.) * swe_fcst * (1. - Observations(ind_obs(1))%obs/beta_threshold)`
That is, replace `max_incr_swe` with `swe_fcst`, which makes the increment should be proportional to `swe_fcst`.
After realizing this, I then checked the first draft and the final published version of Toure et al 2018, both of which also use `swe_fcst` (_SWE_m_, which is never defined...) and not `max_incr_swe`.
Note that because `asnow_obs<beta` per the elseif condition, we have `-swe_fcst <= swe_incr < 0`.
Does this make sense? 
Also, it's not clear to me immediately what the value of `beta` should be.  This requires more thought and (probably) data.
cc: @lcandre2 @gmao-qliu @amfox37

lcandre2: Thanks @gmao-rreichle. 

Replacing max_incr_swe makes sense. I know we went back and forth on this previously because using max_incr_swe is what is needed to reproduce Figure 2 and it seemed odd that we could remove more snow than could be added (since max_incr_swe remains in the first if statement). But, a higher rate of removal during assimilation makes sense if MOD10C1 overestimates snow presence. From a quick lit search, this seems to be the case, at least for North America. Globally, this isn't so clear, so I'll look into it more.

Beta is (unclearly) related to the difference between the model and MODIS SCF, so I'll come up with a clearer way to define this and other parameters since they will need to be considered when switching satellite platforms. 

@gmao-rreichle @gmao-qliu @amfox37

gmao-rreichle: @amfox37 : The large negative increments make sense because a zero snow cover obs should be reasonably reliable.  If MODIS doesn't see snow, there shouldn't be any in the model.  Whether we should use WEMIN (13 mm nowadays) or SWE_fcst is up for discussion.  The graphic in Toure et al does not really show increments for SWE_fcst>WEMIN.  In case SWE_fcst>WEMIN and SCF_obs=0, we might as well remove all of the snow rather than just the first WEMIN kg/m2.
Re. the second part of the conditional:  I think I removed this in my attempts to simplify the implementation because the opening if condition already ensures that SCF_fcst > SCF_obs*alpha if we get into the elseif statement.  That is, the second part of the conditional should be redundant.  Or did I get this wrong?

lcandre2: After revisiting Toure et al., I am convinced that max_swe_incr is used when Modis obs are less than beta. F2 has an increment range of -5 to 5 the DA-OL figures do not visually display a difference beyond 5 mmwe (even though the color bar goes to 6 and in the sensitivity analysis, there is the following statement: " ... it was concluded that max_swe_incr = 5 mm is a reasonable approximation for the maximum amount of snow that can be added to (**or removed from**) a catchment during the ... rule-based update procedure". But, now I think that what is plotted/used in Toure et al is option 2b below. 

max_swe_incr cannot be used alone during snow removal -- it needs to be scaled with the fcstSCF on line 4552. And honestly, I don't know why it isn't in the code, probably a mistake on my part. This would also pertain to using swe_min.

I think there are two, possibly 3 options: 
 1. swe_incr = (-1.0) * (fcstSWE) * (1 - obsSCF / beta)
This scales the increment by the model SWE, which produces lower negative increments when the SCF is below 1, but can produce really high negative increments if there is full, very deep snow cover.

**2a.** swe_incr = (-1.0) * (swe_min * fcstSCF ) * (1 - obsSCF / beta)
In this case, I've included the forecast SCF. In the first option, the results are the same at SCF>1, but when fcstSCF = 1 and ObsSCF = 0, the maximum negative increment would be capped at swe_min.

OR
**2b.** swe_incr = (-1.0) * (max_swe_incr * fcstSCF ) * (1 - obsSCF / beta)
Here, we could use max_swe_incr instead of min_swe. This would essentially lighten the touch of the negative increment. But, it would probably be best to determine the value of max_swe_incr with a sensitivity study.

Finally, BETA. 40% was chosen because the difference between the model and obs SCF rarely exceeds 40%. I am actually looking into this and coming up with a more precise way to determine this. Same with ALPHA.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,12 +4465,199 @@ subroutine cat_enkf_increments(                               &
...
+                
+                tmp_obs = sum(Observations(ind_obs(1:N_selected_obs))%obs)
+                
+                tmp_obs = tmp_obs/real(N_selected_obs)
+             
+             else
+                
+                tmp_obs = Observations(ind_obs(1))%obs
+                
+             end if
```
gmao-rreichle, line 4532: If statement introduced in the hope of maintaining 0-diff.  Needs cleanup later.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,12 +4465,199 @@ subroutine cat_enkf_increments(                               &
...
+       ! ensure that max SWE increment parameter is less than WEMIN; larger increments make no sense because
+       ! at SWE=WEMIN, the tile is fully snow covered (asnow=1)
+
+       if (SCF_ANA_MAXINCRSWE>WEMIN)  call ldas_abort(LDAS_GENERIC_ERROR, Iam, 'must use SCF_ANA_MAXINCRSWE<=WEMIN')
+       
+       ! get target for snow layer thickness (as used in Catchment over land tiles)
+       
+       targetthick(1)        = CATCH_DZ1MAX
+       
+       targetthick(2:N_snow) = 1./(N_snow-1.)
```
gmao-rreichle, line 4483: Calculation of targetthick(2:N_snow) here is potentially non-0-diff vs. the previously hardcoded DZMAX.  If testing of SCF DA runs returns non-0-diff results, change temporarily to `targetthick(2:N_snow)=0.5`

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -4584,6 +4585,534 @@ subroutine read_obs_SMOS( date_time, N_catd, this_obs_param,      &
...
   end subroutine read_obs_SMOS
 
+
+  ! *****************************************************************
+  
+  subroutine read_MODISscf_hdf( N_files, date_time, N_data, fnames, &
+       MODIS_lon, MODIS_lat, MODIS_SCF)
+    
+    ! Purpose: read preprocessed (renamed) snow cover data from daily MODIS Terra MOD10C1, version 6.1 (https://nsidc.org/data/mod10c1/versions/61) 
+    !   - Data currently located at /discover/nobackup/projects/S2SHMA/MODIS/MOD10C1_V61/ (2010-2022)
```
gmao-rreichle, line 4595: Not sure what "preprocessed" means here.   The files in the Discover dir appear to be the original MOD10C1.*.061.* files except for the removal of the portion of the file name that  reflects the granule production date.   Might need a process like that for SMAP Tb obs, whereby a text file with a list of file names is generated outside of GEOSldas and then read into GEOSldas.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -4584,6 +4585,534 @@ subroutine read_obs_SMOS( date_time, N_catd, this_obs_param,      &
...
+    end if
+    
+    ! Allocate SCF and QC vectors
+
+    allocate(SCF_raw(    N_data))    
+    allocate(CI_Index(   N_data))
+    allocate(Snow_QA(    N_data))
+    allocate(Cloud_Index(N_data))
+
+    ! read hdf data into arrays, concatenate data from N_files files   !?!?!?!?! NOT HAPPENING...
```
gmao-rreichle, line 4736: Not sure if we need the functionality to read more than one file at a given analysis time.  If so, the current reader does not provide this functionality.  If it is not needed, then the reader needs to be cleaned up so it's obvious that only one file is read.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,12 +4465,199 @@ subroutine cat_enkf_increments(                               &
...
+                      
+                      ! snow in forecast and analysis, derive properties of analysis snow from properties of forecast snow
+                      
+                      tmp_wesn(kk,n_e,isnow) = cat_progn(kk,n_e)%wesn(isnow) * swe_ratio
+                      tmp_htsn(kk,n_e,isnow) = cat_progn(kk,n_e)%htsn(isnow) * swe_ratio  ! What if fcst snow is at 0 deg C and contains liquid water?  TBD reichle 20221007, could update as:
+                      !tmp_htsn(kk, n_e, isnow)   = (cpw * tpsn(kkm n_e, isnow)  - fices(kk, n_e, isnow) * MAPL_ALHF) * tmp_wesn(kk, n_e, isnow)
+                      
+                      if (asnow_ana < 1.) then
+                         
+                         tmp_sndz(kk,n_e,isnow) = cat_progn(kk,n_e)%sndz(isnow)          ! snow depth remains constant b/c swe_ana<1.
```
gmao-rreichle, line 4609: @lcandre2, @amfox37, @gmao-qliu: I'm wondering if we need another if condition here.  If asnow_ana<1 **and** asnow_fcst<1, then keeping the snow depth constant as coded makes sense.  I'm not sure what we would want to do for asnow_ana<1 **and** asnow_fcst==1.  In this case, we had full snow coverage before the analysis and SWE potentially well above 13 kg/m2, with a correspondingly deep snow pack.  In this case, removing SWE and simultaneously keeping the depth constant would presumably reduce the snow density (potentially even below the min value), but we want to keep the density constant.  I'm struggling to figure out what we would want to do instead. 
 Here's my best guess late on a Fri afternoon:
```
if (asnow_ana < 1.) then
  if (asnow_fcst < 1.) then
    tmp_sndz(kk,n_e,isnow) = cat_progn(kk,n_e)%sndz(isnow)          ! snow depth remains constant b/c swe_ana<1.
  else
    tmp_sndz(kk,n_e,isnow) = tmp_wesn(kk,n_e,isnow) / ( cat_progn(kk,n_e)%wesn(isnow) / cat_progn(kk,n_e)%sndz(isnow) ) / asnow_ana
  end if
else
  ! compute analysis snow depth from (layer-specific) forecast snow density
  tmp_sndz(kk,n_e,isnow) = tmp_wesn(kk,n_e,isnow) / ( cat_progn(kk,n_e)%wesn(isnow) / cat_progn(kk,n_e)%sndz(isnow) )
end if
```
**If** this makes any sense at all, then __maybe__ we could merge the `else` statements as follow (not 100% sure of the logic...):

```
if ( (asnow_ana < 1.) .and. (asnow_fcst < 1.) ) then
  tmp_sndz(kk,n_e,isnow) = cat_progn(kk,n_e)%sndz(isnow)          ! snow depth remains constant b/c swe_ana<1.
else
  ! compute analysis snow depth from (layer-specific) forecast snow density
  ! (need to divide by asnow_ana to get sndz within the snow-covered portion in case asnow_ana<1)  
  tmp_sndz(kk,n_e,isnow) = tmp_wesn(kk,n_e,isnow) / ( cat_progn(kk,n_e)%wesn(isnow) / cat_progn(kk,n_e)%sndz(isnow) ) / asnow_ana

end if
```
In any case, this probably requires another round of thinking.

lcandre2: @gmao-rreichle, I agree that maintaining the snow depth in the situation where fcst=1 and ana<1 could lead to physically unrealistic (low) densities and the fix should be similar to the situation where fcst<1 and ana=1. Your proposed solution makes sense to me -- I would have forgotten the inclusion of asnow_ana.  The simplification is logical to me, in the fcst<1 & ana=1, the last term in depth calculation would be 1, making it identical to fcst=1 & ana<1. No point in having two else statements when 1 works.

lcandre2 (approved):

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
lcandre2: I am making and testing this modification now.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,6 +4465,193 @@ subroutine cat_enkf_increments(                               &
...
+                   ! ADD SNOW:    Forecast SCF is less than observed SCF (after "bias" adjustment with alpha)
+                   
+                   swe_incr(kk,n_e) = SCF_ANA_MAXINCRSWE * (tmp_obs - asnow_fcst/SCF_ANA_ALPHA)
+                   
+                elseif (tmp_obs .lt. SCF_ANA_BETA) then
+                   
+                   ! REMOVE SNOW: Simulated SCF is greater than observed SCF (after "bias" adjustment)
+                   !              and observed SCF is less than beta threshold
+                   
+                   swe_incr(kk,n_e) = (-1.) *  SCF_ANA_MAXINCRSWE * asnow_fcst * (1. - tmp_obs/SCF_ANA_BETA)
```
gmao-rreichle, line 4555: I noticed that there's a minor blemish with the revised version of Toure et al 2018 here.  Specifically, swe_incr has a small discontinuity across the line that separates the negative and positive increments (line through origin with slope alpha).  I don't think this necessarily needs to be changed, but we may want to discuss.  To illustrate, here's a plot with the parameters alpha=0.6, beta=0.55, and maxsweincr=5.0 (as currently hardwired in clsm_ensupd_glob_param.F90):
tmpToureEtAl2018eq1

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,6 +4465,217 @@ subroutine cat_enkf_increments(                               &
...
+                         
+                         ! compute analysis snow depth by keeping snow density constant
+                         !
+                         ! in this case, it is possible that either asnow_fcst<1 or asnow_ana<1;
+                         ! when computing density or depth, make sure that SWE value (which is per unit area) is 
+                         ! adjusted to reflect SWE value (per unit area) in the snow-covered fraction of the tile
+                         
+                         ! i) diagnose (layer-specific) forecast snow density 
+                         
+                         snow_dens = ( cat_progn(kk,n_e)%wesn(isnow)/asnow_fcst ) / cat_progn(kk,n_e)%sndz(isnow)
```
gmao-rreichle, line 4635: This calculation now accounts for the possibility that asnow_fcst<1, which I think could be the case here.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,6 +4465,217 @@ subroutine cat_enkf_increments(                               &
...
+                      !
+                      !   call StieglitzSnow_calc_tpsnow(1, cat_progn(kk,n_e)%htsn(isnow), cat_progn(kk,n_e)%wesn(isnow), snow_temp, fice)
+                      !
+                      !   tmp_htsn(kk, n_e, isnow) = (cpw*snow_temp - fice*MAPL_ALHF)*tmp_wesn(kk,n_e,isnow)
+                      !
+                      ! but not sure if the latter equation is correct 
+                      !
+                      ! for now, keep multiplication with swe_ratio:
+                      
+                      tmp_htsn(kk,n_e,isnow) = cat_progn(kk,n_e)%htsn(isnow) * swe_ratio
```
gmao-rreichle, line 4615: The snow heat content update in this case probably requires more thought.   The multiplication with swe_ratio should be fine if there is not liquid water in the snow pack.  It may not be quite right if there is liquid water, but I'm not sure what to do instead.  The comments in the code include a suggestion that may or may not make sense.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -4403,6 +4465,217 @@ subroutine cat_enkf_increments(                               &
...
+                elseif (tmp_obs .lt. SCF_ANA_BETA) then
+                   
+                   ! REMOVE SNOW: Simulated SCF is greater than observed SCF (after "bias" adjustment)
+                   !                and observed SCF is less than beta threshold
+                   
+                   swe_incr(kk,n_e) = (-1.) *  SCF_ANA_MAXINCRSWE * asnow_fcst * (1. - tmp_obs/SCF_ANA_BETA)
+                   
+                else 
+                   
+                   cycle  ! NO CHANGE, skip rest of increment calcs and go straight to next ens member
```
gmao-rreichle, line 4545: Note new "cycle" statement here.  I think this is right, but if the revised code gives very different results, this change should be examined further

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
lcandre2: Just the renaming, but I think we will need to include a filelist because there is a period in Feb 2016 where there are no modis data

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -4584,6 +4585,534 @@ subroutine read_obs_SMOS( date_time, N_catd, this_obs_param,      &
...
+    !   - Daily dataset with spatial resolution of 0.05 deg on CMG grid, missing days 2016 d. 50-58
+    ! 
+    ! Procedures:
+    !   - dtstep_assim and ref_time is restricted to 3 hr and 00z, respectively
+    !   - Generate the latitude and longitude on CMG grid
+    !   - Select longitude band considering MODIS observation time (local time: ~10:30am)     ! HARDWIRED OVERPASS TIME, WHAT ABOUT AQUA !?!?!!?!?!?!?!
+    !
+    ! Table. Longitude band based on the assimilation time step (in UTC) 
+    !-----------------------------------------------------------------------------
+    ! UTC | 0000    | 0300   | 0600  | 0900 | 1200 | 1500   | 1800    | 2100
```
lcandre2, line 4605: There is an error in how the longitude bands are constructed such that the edges are included in two assimilation cycles. I am trying to identify where this happens, but havent quite yet.
image

gmao-rreichle: Thanks, @lcandre2.  I don't think this is worth pursuing right now.  I will rewrite the reader eventually.  I had hoped to get it done before the shutdown, and there's still a sliver of a chance.  If it doesn't happen before Monday, it'll happen after the shutdown.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
lcandre2: If there is liquid water, then the heat content *should* be zero. So if we want to maintain the layer state (fraction of ice and water) and only increase the mass, the ratio should work. However, if we want to add only frozen mass, then I think we would want to calculate the heat content directly. The above equation suggestion follows Stieglitz (if I remember correctly) ... in any case, we would need to import snow layer temperature to calculate directly. More to discuss...

gmao-rreichle: > If there is liquid water, then the heat content should be zero

I don't think this is true. If there is liquid water, the temperature should be 0 deg C.  But there could be a mix of frozen and liquid at 0 deg C, and the heat content of the frozen portion would be a negative number

gmao-rreichle: Changed to determine analysis snow heat content by keeping snow temperature constant (https://github.com/GEOS-ESM/GEOSldas/pull/512/commits/ff473244ae77f4b685bf80ca32aff956e4a7af27)

gmao-rreichle:
@lcandre2, @saraqzhang: 

I updated the MODIS SCF DA branch of the present PR to work with the GCM GC branch of https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/834 (or so I hope).  The latter includes new checks on the snow states after applying the snow increments in the AGCM Corrector segment.  It also includes the new FICE[n] exports from Catchment.  (These exports aren’t yet available in Surface or in the LDAS Ens GridComp.)  See also note below about applying the snow increments in the AGCM.

In the MODIS SCF analysis update, I changed the calculation of the analyzed HTSN so the snow temperature is kept constant (as opposed to using “swe_ratio”). 

I ran the “nightly” GEOSldas tests for Intel and got 0-diff.  This just establishes that I didn’t break what’s in the “develop” branch, but it does **not** test the new MODIS SCF DA functionality.

I have not yet started rewriting the MODIS SCF reader.  The current one has lots of problems, but it is going to have to do for now.

For now, let’s see if the LADAS works with the improvements in the calculation of the LDAS increments and with the new snow checks in the AGCM.  When you're back in the office, please give the branch a try. 

---

Re. https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/834, I did **not** yet implement any of the options that we discussed on Friday:

Option 1: Use only the total SWE increment from LDAS increments file and determine layer-specific SWE, HTSN, and SNDZ updates for the AGCM as in update_type 11 of the LDAS.

Option 2: Apply positive SWE increments by increasing snowfall.  (Besides the complications we already discussed, I realized that this is tricky because we would have to persist the increased snowfall beyond just one 7.5-minute time step, and it is unclear how long the snowfall should last.)

Option 3: Replace the AGCM forecast snowpack with the LDAS analysis snowpack.  

cc: @amfox37 @gmao-qliu

lcandre2:
@gmao-rreichle, @saraqzhang , the SCF DA branch works (builds and runs) with the modifications to GEOSgcm_GridComp.

gmao-rreichle:
All GEOSldas nightly 0-diff tests passed after most recent commit.

lcandre2:
@gmao-rreichle, the newest iteration with Aqua and Terra runs cleanly.

biljanaorescanin (approved):

Issue #513, opened 2022-01-14
GEOSldas version info in GEOSldas/CMakeLists.txt


gmao-rreichle:
@mathomp4, @weiyuan-jiang:
I just noticed that GEOSldas version info is encoded in Line 7 of GEOSldas/CMakeLists.txt:
https://github.com/GEOS-ESM/GEOSldas/blob/8ef1476258198752af063a4b7c49f90c8b4c03da/CMakeLists.txt#L7
The most recent GEOSldas release is v17.10.0, and we did not update CMakeLists.txt to match the increased "minor" version number.
Is the GEOSldas version info in CMakeLists.txt really necessary?  It's difficult to remember to edit this file prior to issuing a release.  This is even more pertinent now that we're planning to increment the "minor" version number much more frequently.
Please advise, thanks.

tclune:
@gmao-rreichle We too had/have trouble remembering to update this prior to commits.  Have gotten better over time, but still miss.    As @mathomp4 explained, at the fixture level there are few if any consequences to this, so feel free to rip it out. (If we are wrong, and CMake requires something - make it obviously bogus like `0.0.0`)    For libraries though this information is very important as upstream packages may require a certain minimum version or change logic depending on what version is available.   From that perspective it is most important to get the "major" version correct, and maybe "minor" on occasion.    But even "patch" changes for bug fixes could in theory be important for upstream projects no know.

It would be nice to somehow enforce that either this entry updates with GitHub releases or GitHub releases "guess" version from this entry.   But enough exceptions, that only good process is possible.    Have a checklist for the things that must be done before each release ...  (And then get someone more reliable than myself to execute said list!)

gmao-rreichle:
Thanks, @mathomp4 and @tclune, for the info.  I'm leaning towards removing the Version entry from CMakeLists.txt for the time being.  
We could probably manage to remember updating CMakeLists.txt with a checklist-based approach.  I'm getting better at editing Changelog.md prior to issuing the Release, so I "only" have to remember to edit CMakeLists.txt along with Changelog.md.
But, as you say, having the version info is not as critical for Fixtures.  Perhaps most importantly, users will edit/modify GEOSldas releases without changing the version info in CMakeLists.txt.  That is, in many instances, the report in the run log wouldn't match the build that users actually run.   
[...]

PR #514, opened 2022-01-19
Remove GEOSldas version info from CMakeLists.txt

gmao-rreichle:
Addresses #513.  Tentatively removing GEOSldas version info from CMakeLists.txt.

The reason for doing this is **not** the burden of maintaining the GEOSldas version info in CMakeLists.txt (although we missed updating the version entry for v17.10.0).  

Rather, the concern is that even if the GEOSldas version info is maintained properly in CMakeLists.txt, in many cases the version info would at best provide a starting point because it cannot reflect changes that users make to GEOSldas before the build. 

Going forward, it may be possible to assemble version info from git as suggested by @tclune:  https://github.com/GEOS-ESM/GEOSldas/issues/513#issuecomment-1015471164

@biljanaorescanin : When you get a chance, please test the PR to make sure we can get away with not having GEOSldas version info in CMakeLists.txt.  If there is a problem, we can restore the version info, set the version to 0.0.0 (something that's obviously bogus), and try again.  

cc: @mathomp4 @weiyuan-jiang

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,6 @@ cmake_policy (SET CMP0054 NEW)
 
 project (
   GEOSldas
-  VERSION 17.9
   LANGUAGES Fortran CXX C)  # Note - CXX is required for ESMF
 
 if ("${PROJECT_SOURCE_DIR}" STREQUAL "${PROJECT_BINARY_DIR}")


biljanaorescanin:
This PR passed all nightly tests.

biljanaorescanin (approved):

mathomp4 (approved):
Seems safe!

gmao-rreichle:
Just for the record, after removing the GEOSldas version info from CMakeLists.txt in this PR, the GEOSldas build now issues a warning (see below).  I suspect we could make the warning disappear by adding a dummy version ("0.0.0").  I don't think that's critical but let me know if you'd rather see a dummy version than nothing at all.

```
CMake Warning at @cmake/@ecbuild/cmake/ecbuild_log.cmake:158 (message):
  WARN - Please set a version in the project statement:

         project( GEOSldas VERSION x.x.x LANGUAGES C CXX Fortran )
```

cc: @weiyuan-jiang @tclune @mathomp4 @biljanaorescanin

tclune:
Ideally we could suppress the warning, but I don't immediately see a way to do that.    I'm a bit surprised that _ecbuild_ is where the warning comes from as `project` is a CMake command.

But a version of 0.0.0 is preferable to the warning, IMO.  Still obvious that it is meaningless, but does not create noise in the cmake step.     We already have more warnings there than @mathomp4 and I would like.

mathomp4:
Hmm. Yeah. ecbuild will do that, it looks like. And I'm pretty sure `ecbuild_declare_project()` is required.

Indeed, if we ever update to be based on the mother ecbuild, [I think it is *required*](https://github.com/ecmwf/ecbuild/blob/77ad783236c72e66c449a8b58000b751807c80f3/cmake/ecbuild_declare_project.cmake#L92-L106) as the else path that leads to that `ecbuild_warn` in our ecbuild, is gone and only an `ecbuild_critical` remains.

PR #517, opened 2022-01-27
GEOSldas changes prompted by cleanup of catch_constants in GEOSgcm GC

gmao-rreichle:
Matches https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/525

Merge this PR once the above GEOSgcm GC PR has been merged.

**Before merge, restore "develop" version of** `components.yaml` **on this feature branch.** 

Preliminary testing is in progress.  UPDATE: Intel tests with standard optimization passed.  Full set of tests to be conducted by @biljanaorescanin when PR is "ready for review".

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -35,6 +35,5 @@ MAPL:
 GEOSgcm_GridComp:
   local: ./src/Components/GEOSldas_GridComp/@GEOSgcm_GridComp
   remote: ../GEOSgcm_GridComp.git
-  branch: develop  
+  branch: feature/rreichle/cleancatchconstants
   sparse: ./config/GEOSgcm_GridComp_ldas.sparse
-  develop: develop
--- a/components.yaml
+++ b/components.yaml
@@ -35,5 +35,6 @@ MAPL:
 GEOSgcm_GridComp:
   local: ./src/Components/GEOSldas_GridComp/@GEOSgcm_GridComp
   remote: ../GEOSgcm_GridComp.git
-  branch: feature/rreichle/cleancatchconstants
+  branch: develop  
   sparse: ./config/GEOSgcm_GridComp_ldas.sparse
+  develop: develop
--- a/src/Applications/LDAS_App/preprocess_ldas_routines.F90
+++ b/src/Applications/LDAS_App/preprocess_ldas_routines.F90
@@ -60,9 +60,11 @@ module preprocess_ldas_routines
        N_force_pert_max
   
   use catch_types,                     ONLY:   &
-       cat_param_type,                         &
-       N_gt
-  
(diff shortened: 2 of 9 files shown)


gmao-rreichle:
@weiyuan-jiang: 
In putting together this PR, I noticed that two constants (`N_gt` and `N_snow`) that originate in the GEOSgcm_GridComp file: 
`GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/GEOSland_GridComp/Shared/catch_constants.f90`
are also hardwired in a GEOSldas file: 
`./src/Components/GEOSldas_GridComp/Shared/catch_types.F90`

It looks like at some point we had a "use..." statement in `catch_constant.f90` that is now commented out:
https://github.com/GEOS-ESM/GEOSldas/blob/81a4232b35e7697e9061ee5b56df6ef5878b1a7e/src/Components/GEOSldas_GridComp/Shared/catch_types.F90#L24
I tried uncommenting this statement but the GEOSldas build fails because for some reason `catch_types.F90` is compiled before `catch_constants.f90`.

Also, there is another GEOSldas file (`LDAS_ensdrv_mpi.F90`) in the same dir that also uses the same two constants (and currently pulls them from `catch_types.F90`):
https://github.com/GEOS-ESM/GEOSldas/blob/81a4232b35e7697e9061ee5b56df6ef5878b1a7e/src/Components/GEOSldas_GridComp/Shared/LDAS_ensdrv_mpi.F90#L48

Can we force CMake to compile `catch_constants.f90` before `catch_types.F90`?  Or would we need to change more than the order in which the files are compiled?  It would be good to avoid the duplicate definitions of these constants.

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang:
Scott just merged the associated GEOSgcm_GridComp PR:  https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/525
This means that we need to merge the present GEOSldas PR to keep the GEOSldas tests working. 
I had tested the branch initially but not since it was finalized. 
There isn't enough time today to test the branch before the nightly LDAS tests kick off.  We could (1) merge the branch and let the automated nightly tests do the verification.  Or we could (2) let the automated nightly tests fail today, do a manual test, and merge the branch tomorrow.  
Since I initiated the PR, I can't approve/merge.  I'll leave it up to you and your availability to tackle this.  Either way is fine with me. 
cc: @mathomp4

biljanaorescanin:
I will merge it since as you said it would fail anyway. And there is not enough time to run tests now.

biljanaorescanin (approved):

PR #523, opened 2022-02-03
change MAPL_ENABLE_BOOTSTRAP to YES

weiyuan-jiang:
set MAPL_ENABLE_BOOTSTRAP to YES by default

### DIFF
--- a/src/Applications/LDAS_App/GEOSldas_LDAS.rc
+++ b/src/Applications/LDAS_App/GEOSldas_LDAS.rc
@@ -214,7 +214,7 @@ DYCORE: none
 
 # ---- For MAPL_RestartOptional
 #
-MAPL_ENABLE_BOOTSTRAP: NO
+MAPL_ENABLE_BOOTSTRAP: YES
 
 #----- Write restart or checkpoint by oserver
 #


gmao-rreichle (approved):

gmao-rreichle:
@mathomp4, @weiyuan-jiang:  
1. Now that the GEOSldas default for MAPL_ENABLE_BOOTSTRAP is YES, the "utility.py" script of the LDAS nightly tests should be updated.  Specifically, the "utility.py" script should **not** edit MAPL_ENABLE_BOOTSTRAP.
2.  Are there other GEOSldas defaults that are modified (through "utility.py" or otherwise) in the LDAS nightly test script?  We should have the GEOSldas defaults be as consistent as possible with what we're testing.   

cc: @biljanaorescanin

mathomp4:
> @mathomp4, @weiyuan-jiang:
> 
> 1. Now that the GEOSldas default for MAPL_ENABLE_BOOTSTRAP is YES, the "utility.py" script of the LDAS nightly tests should be updated.  Specifically, the "utility.py" script should **not** edit MAPL_ENABLE_BOOTSTRAP.
> 2. Are there other GEOSldas defaults that are modified (through "utility.py" or otherwise) in the LDAS nightly test script?  We should have the GEOSldas defaults be as consistent as possible with what we're testing.
> 
> cc: @biljanaorescanin

@gmao-rreichle Are you talking about lines like this in the exec files:
```
input/LDAS/global.model.exec
284:MAPL_ENABLE_BOOTSTRAP: YES
```
Otherwise, the scripting doesn't actually alter bootstrapping I think.

As for your other question, I think the only edit I can see is to `LDAS.rc` if you are running with Open MPI (aka GNU). For GNU tests, it will set `WRITE_RESTART_BY_OSERVER: YES` for any case with GNU due to, well, GNU+Open MPI weirdness with restarts. Maybe this isn't needed anymore?

gmao-rreichle:
@mathomp4: Thanks for the info.  I'm a bit confused where MAPL_ENABLE_BOOTSTRAP actually turns into YES in the LDAS nightly tests.  Earlier today, @weiyuan-jiang emailed the following:
```
In our night tests, the 'NO' is replaced with 'YES' by a utility.py

  for line in lines:
        if ('MAPL_ENABLE_BOOTSTRAP' in line) and ('YES' not in line):
            line = line.replace('NO', 'YES')
        gout.write(line);
```
If "utility.py" is no longer touching MAPL_ENABLE_BOOTSTRAP (perhaps Weiyuan looked at an older version of the test scripts), then we should probably remove the MAPL_ENABLE_BOOTSTRAP line from the *.exec configuration input files. With the GEOSldas default set to YES now, these shouldn't be needed.

Thanks for the info about `WRITE_RESTART_BY_OSERVER: YES` in GNU.  I can't say if this is still needed.  @weiyuan-jiang, can you please try and see if GNU still requires special treatment?  If yes, so be it.  But if we can drop this special case, we probably should.

weiyuan-jiang:
> As for your other question, I think the only edit I can see is to `LDAS.rc` if you are running with Open MPI (aka GNU). For GNU tests, it will set `WRITE_RESTART_BY_OSERVER: YES` for any case with GNU due to, well, GNU+Open MPI weirdness with restarts. Maybe this isn't needed anymore?

I prefer to leave "YES" there. It give us a test for  oserver

gmao-rreichle:
> I prefer to leave "YES" there. It give us a test for oserver

Ok with me.  

Re. Weiyuan's note:  I think the python script is called `utils.py`, and it touches MAPL_ENABLE_BOOTSTRAP only in a function called bootstrapAGCM, so not related to the LDAS tests. 

This leaves **one action item for** @mathomp4:  Please remove the now-redundant  `MAPL_ENABLE_BOOTSTRAP: YES` lines from the *.exec config input files in the LDAS tests.   Thanks!

weiyuan-jiang:
Indeed, it is set in *.exec input files not in the script.

mathomp4:
Okay. I'll remove them now and we'll see how everything goes tonight!

PR #526, opened 2022-02-09
Script to compress bit-shaved netcdf output 

gmao-rreichle:
This PR adds a utility to compress bit-shaved netcdf output. 

Original version provided by @gmao-qliu, lightly edited by @gmao-rreichle.

@weiyuan-jiang: Please review and add multi-threading capability.


weiyuan-jiang:
@gmao-qliu , I thought we can use the same name as input and output. Can you test my changes?

gmao-qliu:
> 

it's unsafe to use the same name for input and output according to the nco user guide.

weiyuan-jiang:
> it's unsafe to use the same name for input and output according to the nco user guide.

I think nco internally did the same thing as you did.

src/Applications/LDAS_App/util/compress_nc.sh
```diff
@@ -0,0 +1,24 @@
...
+	usage
+	exits;
+fi
+
+# deflation_level 1-9; higher level (e.g. 9) takes longer; recommended: 3 (based on L4_SM) 
+deflate_level=3
+
+for file in $(find $1 -name "*.nc4" -type f); do
+	#echo "Processing $file..."
+	ncks -L $deflate_level -O $file $file &
```
gmao-rreichle, line 22: @weiyuan-jiang:  I'm concerned the ampersand at the end of the `ncks` command will create problems if there are many files in the directory.  If there are, say, 100 nc4 files, then 100 ncks commands would be kicked off nearly instantaneously, which may bog down the login node.  For multi-threading, I think we would need a more sophisticated approach that limits the number of simultaneous ncks jobs to, say, 50% of the number of available processors.  I imagine this isn't trivial, and it may be simpler just to run the ncks jobs sequentially.  Thoughts?

gmao-rreichle:
> > it's unsafe to use the same name for input and output according to the nco user guide.
> 
> I think nco internally did the same thing as you did.

@gmao-qliu: Is it ok with you to use the same file name going into and out of ncks?

gmao-qliu:
> 

It's fine with me. Seems it's only unsafe with the --no_tmp_fi flag. @gmao-rreichle @weiyuan-jiang

src/Applications/LDAS_App/util/compress_nc.sh
weiyuan-jiang: usually if a users has many jobs, he should not do it on login nodes.  If time is not an issue, sequential jobs is totally fine.

gmao-rreichle:
@weiyuan-jiang, @biljanaorescanin:  I've finished my modifications of this PR but I can't approve it because I initiated it.  When you get a chance, please approve and merge, or let me know if you have any concerns.

biljanaorescanin (approved):

PR #531, opened 2022-02-23
bug fix: dimensions of WATERTABLED export in GEOS_EnsGridComp.F90 (MAPL_DimsTileOnly)

weiyuan-jiang:
bug fix.


gmao-rreichle:
Thanks, @weiyuan-jiang, for identifying and fixing this bug.  It slipped through the cracks in 0-diff testing vs. develop because no test was writing out this new export.  And we never ran PEATCLSM with this output in ensemble mode, I think. 
In any case, did you test the fix in the sense that we can now write out the ens_avg WATERTABLED?  Or do we just assume that it'll work?  It should from just looking at the src code.

weiyuan-jiang:
@gmao-rreichle , I found this bug because of Qing's work.  @gmao-qliu please let us know if the fix works for you.

gmao-qliu:
> @gmao-rreichle , I found this bug because of Qing's work. @gmao-qliu please let us know if the fix works for you.

Yes, it works now. @weiyuan-jiang  @gmao-rreichle

gmao-rreichle (approved):

PR #543, opened 2022-03-21
Update components.yaml to match GEOSgcm v10.22.1 (and update CI)

mathomp4:
This PR updates ESMA_env, ESMA_cmake, and MAPL in the GEOSldas to match GEOSgcm v10.22.1. The changes are:

* ESMA_env
  * Update to Baselibs 6.2.13 (v3.12.0)
  * Move to have both Python2 and Python3 loaded at the same time (v3.12.0)
  * Minor tweak to `parallel_build.csh` (v3.13.0)
* ESMA_cmake
  * **_Updates for Spack support (v3.11.0)_**
  * Preliminary M1 support (v3.12.0)
* MAPL
  * well...A lot of bug fixes (see https://github.com/GEOS-ESM/MAPL/blob/main/CHANGELOG.md)

Of these, the ESMA_cmake is needed for build issues as GMAO_Shared `main` needs this version.

Note that there is a chance the MAPL update could be sort-of non-zero-diff. The issue is that Ops found some lats and lons in history output were of the sort of "6.500000000001" instead of "6.5". This MAPL update "fixes" that, and comparators like `nccmp` will see this as a difference. The *data* is zero-diff, but the lats and lons are not.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 orbs:
-  circleci-tools: geos-esm/circleci-tools@0.11.0
+  circleci-tools: geos-esm/circleci-tools@0.13.0
 
 workflows:
   build-test:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,27 +1,7 @@
 version: 2.1
 
-executors:
-  gfortran-large:
-    docker:
-      - image: gmao/ubuntu20-geos-env-mkl:v6.2.8-openmpi_4.0.6-gcc_11.2.0
-        auth:
-          username: $DOCKERHUB_USER
-          password: $DOCKERHUB_AUTH_TOKEN
-    environment:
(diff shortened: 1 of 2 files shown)


biljanaorescanin:
Tests globalcs, aggglobalcs and gnuglobalcs fail for model and assim runs, for both collections ("lnd" and "lfs" ). 
If we compare these tests without lat/lon files are identical. ( nccmp -dfsB -x lat,lon )

gmao-rreichle:
> Tests globalcs, aggglobalcs and gnuglobalcs fail for model and assim run comparison, for both collections ("lnd" and "lfs" ). If we compare these tests without lat/lon files are identical. ( nccmp -dfsB -x lat,lon )

How about the other tests (conus, global, cnclm[xx])?  I assume you're implying that the other tests passed?

gmao-rreichle (approved):

mathomp4:
I know it's merged, but I think that makes sense that only the CS tests showed a difference.

PR #565, opened 2022-06-16
Update ESMA_cmake, ESMA_env, MAPL

mathomp4:
This PR updates the GEOSldas to use the latest versions of three base components:

* ESMA_env (update to Baselibs 6.3.1)
* ESMA_cmake (support for M1, **_change in GNU Release options_**, updates for handling ESMF in builds)
* MAPL (latest stable version)

Of all these changes, the one that might matter is the update to the GNU Release options for GCC 12 support. This is a non-zero-diff change. It should be the same world, but slightly different optimization target (Haswell instead of Westmere)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,5 +18,5 @@ workflows:
               compiler: [ifort, gfortran]
           baselibs_version: *baselibs_version
           repo: GEOSldas
-          mapldevelop: false
+          mepodevelop: false
           persist_workspace: false # Needs to be true to run fv3/gcm experiment, costs extra


gmao-rreichle:
@mathomp4 : Thanks for putting this together. Do you have an existing suite of the NIGHTLY LDAS tests of this branch for us to look at? 
If not, @biljanaorescanin, please run and inspect the full set of tests when you get a chance.  
In any case, please let me know where the test runs are, and I'll also take a look at the L4_SM-like output from the GLOBAL*/assim tests that's not in nc4 format.

mathomp4:
Sigh. I am having a bad day with the CI. Try again.

biljanaorescanin:
@gmao-rreichle as expected most GNU tests failed. They failed for for model run in comparison for "lnd", "lfs" and RST. For assim run comparison fails for "ens_avg.ldas_ObsFcstAna" and all collections for .nc4 format except for "ldas_mwRTMparam" that one passed. 

Differences for all failed cases seem to be very small. 

One interesting thing is that one GNU test did pass "gnudebugconus". @mathomp4 is this because of different optimization? 

Here is path to full log file: 
`/discover/nobackup/borescan/SystemTests/logs/LDAS/2022-06-16/log.17:41:36`

List of tests that failed:
gnuconus         pass  pass  13 min     pass/FAIL          -- / --
gnuglobal         --    --    --        pass/FAIL         pass/FAIL
gnuglobalcs       --    --    --        pass/FAIL         pass/FAIL
gnuglobalcnclm4   --    --    --        pass/FAIL          -- / --
gnuglobalcnclm45  --    --    --        pass/FAIL          -- / --

mathomp4:
> One interesting thing is that one GNU test did pass "gnudebugconus". @mathomp4 is this because of different optimization?

@biljanaorescanin Yes this is expected. I didn't touch the Debug flags and they have no "architecture aware" flag in them. My guess is GNU just targets generic processor which is fine for me. Architecture tuning sometimes screws up tracebacks, and I want to keep Debug boring. :)

mathomp4:
One final note is that this isn't changing GCC version yet. I sort of wanted to get all these flags in, see if @gmao-rreichle approves, etc. Since GCC 12 didn't like the old flags, hard to do comparisons with the different compilers.

Once this is accepted, then I'll look to move to GCC 12 in my tests. I think from my testing GCC 12 is zero-diff to 11, but I'll make sure to test before I definitively say that.

mathomp4:
> @mathomp4, please go ahead and merge if you're ready to do so.

Woot! I don't get to press the button in GEOSldas very often. Merging!

(NOTE: I might also re-run all the GNU tests myself just so the Baselines are good for tonight (otherwise the weekend will be *full* of failures!)

PR #548, opened 2022-04-04
sample files for  ladas setup 

saraqzhang:
a set of sample expinp  and HISTORY.rc  as reference for user to setup of a land-atmos coupled experiment.

PR #546, opened 2022-03-31
add utility codes for generating perturbed/recentered ensemble forcing 

saraqzhang:

PR #569, opened 2022-07-08
Reorg of LDAS_App and additional LADAS-related utilities

gmao-rreichle:
- reorganizes `src/Applications/LDAS_App/` 
- adds matlab routines for generation of Tb scaling parameters 
- adds LADAS-related utilities and sample config files, incl. option for ensemble forcing of LDAS coupled to central simulation in 4dHybridEnVar
- supersedes #546 and #548 

cc: @saraqzhang @gmao-qliu


src/Applications/LDAS_App/ens_forcing/enpert_forc.csh
```diff
@@ -0,0 +1,14 @@
+#!/usr/bin/csh -f 
+
+## env : ADAS_EXPDIR,  GRID,  NENS, RUNDIR
```
gmao-rreichle, line 3: @saraqzhang, this script needs a brief statement of how it fits into GEOSldas and what it does.  Something like:

```
# Script to create re-centered ensemble land forcing (lfo) files, typically used in LADAS:
# 1. Regrids coarse-resolution lfo files from the ADAS atmospheric ensemble to the higher resolution 
#   of the single-member central (or deterministic) simulation.
# 2. Computes perturbations from regridded lfo files.
# 3. Applies perturbations to lfo files from central simulation. 
#
# Requires environment variables ADAS_EXPDIR, GRID, NENS, RUNDIR
```

src/Applications/LDAS_App/ens_forcing/enpert_forc.csh
```diff
@@ -0,0 +1,14 @@
...
+
+set force_cntr =  "${ADAS_EXPDIR}/recycle/holdforc"
+set force_orig =  "${ADAS_EXPDIR}/atmens"   
+set force_rgd = "${ADAS_EXPDIR}/atmens/rgdlfo" 
+set outgrid =  "${GRID}"
+mkdir $force_rgd
+
+$RUNDIR/regrid_forc.csh  $force_orig $force_rgd $outgrid
```
gmao-rreichle, line 11: @saraqzhang, why is `regrid_forc.csh` in $RUNDIR?  Shouldn't this live in `./bin` ?  Any scripts needed by ldas_setup and lenkf.j should be placed into `./bin`.  This probably requires editing `../CMakeLists.txt`.

src/Applications/LDAS_App/ens_forcing/enpert_forc.csh
```diff
@@ -0,0 +1,14 @@
...
+
+set force_cntr =  "${ADAS_EXPDIR}/recycle/holdforc"
+set force_orig =  "${ADAS_EXPDIR}/atmens"   
+set force_rgd = "${ADAS_EXPDIR}/atmens/rgdlfo" 
+set outgrid =  "${GRID}"
+mkdir $force_rgd
+
+$RUNDIR/regrid_forc.csh  $force_orig $force_rgd $outgrid    
+rm -rf  $force_orig/tmp* 
+python  $RUNDIR/ensemble_forc.py $force_rgd $force_cntr $NENS
```
gmao-rreichle, line 13: @saraqzhang: Same question, why is `ensemble_forc.py` in $RUNDIR?

src/Applications/LDAS_App/CMakeLists.txt
```diff
@@ -16,11 +16,11 @@ ecbuild_add_executable (
 
 ecbuild_add_executable (
    TARGET mwrtm_bin2nc4.x
-   SOURCES mwrtm_bin2nc4.F90
+   SOURCES util/inputs/mwrtm_bin2nc4.F90
    LIBS GEOSlandassim_GridComp)
    
 install(
-   PROGRAMS ldas_setup process_hist.csh process_rst.csh util/average_ensemble_forcing.py
+   PROGRAMS ldas_setup process_hist.csh process_rst.csh ens_forcing/average_ensemble_forcing.py
```
gmao-rreichle, line 23: CMakelists.txt should need more edits.  Some of the new scripts in ./ens_forcing should also be copied to ./bin.

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.atmensC90
```diff
@@ -0,0 +1,624 @@
+####################################################################################
```
gmao-rreichle, line 1: @saraqzhang: If we include a full copy of this file (and the corresponding "centralC180" file) in the repo, it will be incredibly difficult to keep these sample files in synch with future changes in the exeinp contents. 

Perhaps we can strip down the sample files to only the LADAS-relevant and non-obvious parameter settings, without all the comment lines.  I'm thinking of MET_PATH, MET_TAG, LADAS_COUPLING, ADAS_EXPDIR, MET_HINTERP, LANDASSIM_DT, LANDASSIM_T0, FIRST_ENS_ID, ENSEMBLE_FORCING, and JOB_SGMT.   Not sure about the model config parameters, perhaps we need CHOOSEMOSFC.  Basically, anything that isn't obvious (such as resolution, experiment ID, experiment domain, start/end times, etc) or matches the default.

Alternatively, we could remove these files from the branch and just put in the GEOSldas Wiki. 

PS: Where does NUM_LDAS_ENSEMBLE come from?  Is this connected to ADAS ensemble size through the fvsetup script?

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.centralC180
```diff
@@ -0,0 +1,623 @@
+####################################################################################
```
gmao-rreichle, line 1: Same comment as for `exeinp.txt.atmensC90` above.
If we don't include the resolution in the sample (or use a resolution placeholder string), then the sample config files can be labeled more simply and more concisely, e.g., `exeinp.txt.Hy4dEnVar.atmens` and `exeinp.txt.Hy4dEnVar.central`.  This clarifies that the samples are not for the 3dVar system.  This info should also be included as a comment at the top of the exeinp.txt sample files. 
Also, we may need to include more info about how the atm ens is used in these examples.  This is the latest config that uses the atm ens to create ensemble forcing for the LDAS associated with the central member, right?

src/Applications/LDAS_App/ens_forcing/enpert_forc.csh
saraqzhang: agree.

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.atmensC90
saraqzhang: If we have a strip-down version of  expinp.txt,  it can be put in the the GEOSldas Wiki along with ReadeMe_LADAS. On the other hand,   user cannot use it directly in ldas setup, we need to instruct user to run ldas_setup sample --exeinp  first to generate a complete expinp txt file, then to edit it according  to the items in the strip-down version.

saraqzhang: For  the case coupled with atmens, NUM_LDAS_ENSEMBLE  needs to be the same as in ADAS Nens.   For the case coupled with central, it is a default (24) in ldas.

saraqzhang: similar situation for batinp.txt,  we need to instruct user to run ldas_setup  sample --batinp first to get a file to work with.

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.centralC180
saraqzhang: these are good file names with clear information on coupling targets. But we need to instruct user on how these files are used  ( as reference to edit the complete expinp file generated by ldas_setup sample --expinp) . 
We can add some info on the connection of ldas to atm ens , particularly ens forcing and ens incr ( same  Nens on both sides) .

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.atmensC90
gmao-rreichle: > If we have a strip-down version of expinp.txt, [...] user cannot use it directly in ldas setup, we need to instruct user to run ldas_setup sample --exeinp first to generate a complete expinp txt file, then to edit it according to the items in the strip-down version.

I think the stripped-down version of the exeinp file for LADAS is the better path because it will minimize maintenance going forward.  Unless we are very diligent about keeping up the LADAS-specific exeinp.txt sample files, the user will not be able to use them directly in ldas_setup either.  For the LADAS, the instructions should follow the general guidance of setting up LDAS (incl. the creation of a sample exeinp file through ldas_setup), and minimize the LADAS-specific instructions to only the differences or custom settings that apply to the LADAS.  Please put together a stripped-down version of the exeinp settings for LADAS.

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.centralC180
gmao-rreichle: Ok, thanks.  Please go  ahead and implement these changes.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_GEOSldas.m
```diff
@@ -0,0 +1,767 @@
...
+
+for day = 1:days_in_month( 2014, month) %2014 = random non-leap year
+
+    if count < w_days
+        count = count + 1;
+    else
+        count = w_days;
+    end
+
+  for seconds_in_day = 0:dt_assim:(86400-1)
```
gmao-rreichle, line 339: Since GEOSldas v17.9.3, we have two config variables for the assimilation time stepping:  "LANDASSIM_DT" and "LANDASSIM_T0".
For our usual offline runs, incl. SMAP L4_SM, we have been using LANDASSIM_T0=000000, but for the coupled land-atm DA system (LADAS), we are moving towards LANDASSIM_T0=013000.  It will be a while before we have enough output from LADAS experiments to compute obs and model stats.  Nevertheless, it's probably best to add another parameter ("t0_assim") to this matlab script and its driver routine ("run_*.m").

src/Applications/LDAS_App/sample_config_files/LADAS/HISTORY.rc.atmens
```diff
@@ -0,0 +1,1114 @@
+
+VERSION: 1
+EXPID:  ldas1711_C90
+EXPDSC: GEOSldas_output
+EXPSRC: GEOSldas
+
+COLLECTIONS:
```
gmao-rreichle, line 7: @saraqzhang: 
Has this HISTORY.rc file been generated with the help of `generate_catchincr_hist.py`?  If so, please add a comment into the sample HIST rc file that mentions the script. 
Also, the "catch_progn_incr" output is in tile space and the associated HIST.rc entries are therefore independent of the resolution of the LADAS.  Only the "lndfcstana" output is explicitly linked to the LADAS resolution.  It would probably be better to
(i) define "lndfcstana" **before** the long list of "catch_progn_incr" definitions in the sample HIST rc file, 
(ii) add a comment that the "lndfcstana" resolution should be adjusted to be consistent with that of the LADAS, and
(iii) add a comment that the "catch_progn_incr" output is in tile space and its definition is generic for any LADAS resolution.

src/Applications/LDAS_App/sample_config_files/LADAS/exeinp.txt.centralC180
saraqzhang: I just pushed updates.

src/Applications/LDAS_App/CMakeLists.txt
saraqzhang: I just pushed updates in ens_forcing/ .  CmakeLists is modified to install the codes to bin/.

src/Applications/LDAS_App/sample_config_files/LADAS/HISTORY.rc.atmens
gmao-rreichle: Thanks, @saraqzhang!  I'll take another look at the updates tomorrow and might tweak the wording a bit.
Re. (1): Correct, the sample files in this directory should not be installed.  No need for CMake. 
Re. (2): It would have been a bit easier to trace if the old expinp.txt*C90 and *C180 files had been renamed (e.g., `git mv exeinp.txt.atmensC90 exeinp.txt.Hy4dEnVar.atmens`) and then edited.  Since these are new files it's not important here, but in future commits, if we're changing a file name, I would recommend "git mv oldfilename newfilename", rather than deleting the old file outright.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_GEOSldas.m
```diff
@@ -0,0 +1,767 @@
...
+%                to project the Obs (always M36 for SMOS) and Fcst to M36 
+%                M09 obs are administered by tiles (0) that could be anywhere
+%                around the center of the observed pixel (M36)
+%                -----------
+%                | X X X X |
+%                | X O O X |
+%                | X O O X |
+%                | X X X X |
+%                -----------
+% GDL, jan 2014: the above issue that "any" M09 tile in the center (0)
```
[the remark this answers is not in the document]
gmao-qliu: @gmao-rreichle  we'll need to keep "convert_grid". It's for keeping values at the M09 tile (northeastern from center) for a M36 grid. Without it, values at all M09 tiles will be saved and the file size becomes too large. That part of comments is obsolete. We can delete or modify.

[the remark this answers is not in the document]
gmao-rreichle: Thanks for the update, @gmao-qliu.  I updated the comments in `get_model_and_obs_clim_stats_GEOSldas.m`.  Please double-check  https://github.com/GEOS-ESM/GEOSldas/pull/569/commits/44a926f3179597b7face3a394eeca5289533458a

saraqzhang:
the last commit  add the option in lenkf.j.template  for using atmens perturbed forcing in central coupling . also add comment lines in exeinp.txt.Hy4dEnVar.central corresponding to the option.

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -109,13 +109,29 @@ if( ${LSMCHOICE} > 1 ) then # CatchCN Only
...
 
 if ( $LADAS_COUPLING == 1 ) then
-   echo "copying lfo_Nx+- met forcing from $FVWORK to $SCRDIR"
-   /bin/cp -f $FVWORK/*lfo_Nx+-*nc4  $SCRDIR/.
+
+   if ( $ENSEMBLE_FORCING == "YES" ) then
+
+      # create perturbed forcing from central simulation and atm ensemble
+
+      module load python/GEOSpyD/Ana2019.03_py3.7
```
gmao-rreichle, line 122: @saraqzhang: Why do we need to load python here?  I think GMAO_env (currently v4.2.0) loads  python/GEOSpyD/Min4.11.0_py3.9_AND_Min4.8.3_py2.7:
https://github.com/GEOS-ESM/ESMA_env/blob/b4176aa8c4639fa9f3dac7db1a0f061c30225930/g5_modules#L131
Could there be a conflict?

saraqzhang:
checked, all good.

src/Applications/LDAS_App/lenkf.j.template
saraqzhang: I used 3.7 in the offline development and test.  I can test it using 2.7  see if any problem.

gmao-rreichle: There's also "py3.9" in the module that comes with GMAO_env.  I don't understand whether the GMAO_env module is v2 or v3 of python.  Maybe check with Matt?

src/Applications/LDAS_App/util/inputs/obs_scaling_params/run_get_L4_Tb_scale_SMAP.m
```diff
@@ -157,7 +161,7 @@
             
             get_model_and_obs_clim_stats( var_name,                              ...
                 run_months, exp_path, exp_run{n}, domain, start_year, end_year, ...
-                dt_assim, species, obs_param, ...
+                dt_assim, t0_assim, species, obs_param, ...
```
gmao-rreichle, line 164: @gmao-qliu, in this commit, I removed the `_GEOSldas` from the  `get_model_and_obs_clim_stats()`  function.  However, the **4th** call to `get_model_and_obs_clim_stats()` in the run script did **not** have the `_GEOSldas` string in your original code.  Is this a typo?  Or would we need a different version of `get_model_and_obs_clim_stats()` for this case?

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_GEOSldas.m
gmao-rreichle: @gmao-qliu, I took care of this edit, but please double-check.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/run_get_L4_Tb_scale_SMAP.m
gmao-qliu: @gmao-rreichle no need to have a different version. It was not used so I missed editing it with _GEOSldas.

src/Applications/LDAS_App/util/postproc/climatology/Write_L4_sm_clim_stat_bin2nc4.m
```diff
@@ -0,0 +1,208 @@
...
+clear all
+
+write_GEOSldas_tileorder = 0;
+
+time_stamp = '2001_p1_2021_p73';
+specs_tag  = '_W_5p_Nmin_210';
+
+bin_fpath  = '/home/qliu/smap/SMAP_Nature/SMAP_Nature_v10/SMAP_Nature_v10.0/output/SMAP_EASEv2_M09_GLOBAL/stats/cli/';
+
+file_out   = ['L4SM_NRv10.0_cli_',time_stamp,specs_tag,'_sm_wetness_EASEv2_M09.nc4'];
```
gmao-rreichle, line 20: @gmao-qliu: Aren't the percentile values in soil moisture units (m3/m3) rather than dimensionless wetness units?  The gph output variables are sm_rootzone_pctl and sm_profile_pctl.  Using "wetness" in the file name, nc4 attributes, and labels seems confusing.  Maybe this can be cleaned up?  Or am I misunderstanding something?

gmao-rreichle: @gmao-qliu: After further inspection, I think the percentile values that are generated by `get_model_clim_stats.m` are indeed in dimensionless wetness units.  I'm not sure why.  Presumably, the L4 ops python code that converts soil moisture into pctl units must make a matching unit conversion.  
In any case, we should track down the L4 ops python code that uses the clim file and also add it to this repository. 
Ideally, we would also make the computations in `get_model_clim_stats.m` independent of the L4SM-specific post-processing.  The `get_model_clim_stats.m` should really be a generic utility that takes a given field from any GEOSldas output collection and generates the stats and percentile values, without unit conversions.
Maybe I'm still misunderstanding what's going on, but I'm getting a sense that the scripts in their present form could be improved.

src/Applications/LDAS_App/lenkf.j.template
gmao-rreichle: Addressed by @saraqzhang on 28 Jul, see comment below.  Marking conversation as "resolved".

src/Applications/LDAS_App/util/postproc/climatology/Write_L4_sm_clim_stat_bin2nc4.m
gmao-qliu: @gmao-rreichle gph contains wetness output. The python script takes those to compute percentiles. The units conversion in get_model_clim_stat.m is necessary because the Nature Run tavg3h output doesn't include wetness. The script can be modified and used for other variables, but I'm hesitating to make the change now without testing on individual variables with broader value ranges.

gmao-rreichle: @gmao-qliu, thanks for the insights.  I had thought that the L4 ops script uses vol soil moisture as input but couldn't check because of the Discover outage yesterday.  In any case, I cleaned up some more and resurrected the "out_wetness" switch to deal with the unit conversion.  It's minimally invasive and I'm reasonably confident that this will work ok.  Please take a look: https://github.com/GEOS-ESM/GEOSldas/pull/569/commits/3364e14062ffd23a3c8eb71b670b3527c8d80803?diff=split&w=1  (white-space changes hidden).

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_GEOSldas.m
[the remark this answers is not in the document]
gmao-qliu: Checked. Works well.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_GEOSldas.m
gmao-qliu: double-checking done. @gmao-rreichle

src/Applications/LDAS_App/util/postproc/climatology/Write_L4_sm_clim_stat_bin2nc4.m
gmao-qliu: @gmao-rreichle I just verified the changes. It works well.

gmao-rreichle:
Successfully completed LDAS Nightly 0-diff tests for Intel.  Omitting GNU tests because of ongoing switch to gcc12.  Omitting GNU and "aggressive" tests because changes in this PR are very close to being trivially 0-diff, so Intel tests are deemed sufficient.

biljanaorescanin (approved):

mathomp4 (approved):
Approve for cmake!

PR #572, opened 2022-08-26
fix character length

weiyuan-jiang:
Increase the filenames' ( with path) length so the filenames don't get truncated.


gmao-rreichle (commented):
Looks good to me.  Not sure if a formal test was done, or if it's needed.  Could be considered "0-diff trivial", and the build works.  
However, this PR cannot just be merged because this PR has a "conflict" with #569, which moves around some of the utilities that are changed in this PR.  That is, some of the changes here would be lost unless they are also added to #569.  
Merging this PR therefore requires more thought/discussion.

weiyuan-jiang:
I have tested it. It is zero-diff

gmao-rreichle:
I'm hopeful that we can merge #569 very soon.  Let's keep the present PR (#572) on hold for now.  After #569 is merged, I will create a new PR that replicates the present PR on top of (then revised) develop branch.

gmao-rreichle:
Replaced with #574.  Closing.

PR #574, opened 2022-08-31
fix character length in declaration of string variables to avoid truncation of file paths/names

gmao-rreichle:
Increase the character length of string variables for path/file names to avoid truncation.  Before this fix, there were sometimes runtime errors when path/file names got too long. 
The present PR anticipates and requires merging of #569.
The present PR replaces #572, which was on top of "develop" before the #569 merge and would not have preserved some changes because one of the changed utility script was moved.
#572 was tested by @weiyuan-jiang.

### DIFF
--- a/src/Applications/LDAS_App/preprocess_ldas.F90
+++ b/src/Applications/LDAS_App/preprocess_ldas.F90
@@ -18,30 +18,30 @@ program main
   implicit none
   
   character(len=20 ) :: option
-  character(len=200) :: arg1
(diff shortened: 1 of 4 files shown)


gmao-rreichle:
LDAS Nightly 0-diff tests re-run successfully.  Standard Intel tests only because of ongoing change to gcc12.  Intel only is deemed sufficient.

biljanaorescanin (approved):

PR #578, opened 2022-09-09
Update to MAPL 2.25.0

### LINKED ISSUE #577: Build Error in Nightly Runs
All my LDAS builds last night failed with multiple compile errors in `SaltIntSplitter.F90`:
```
/discover/nobackup/mathomp4/SystemTests/builds/LDAS_CONUS/CURRENT/GEOSldas/src/Components/GEOSldas_GridComp/@GEOSgcm_GridComp/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts/SaltIntSplitter.F90(105): error #6460: This is not a component name that is defined in the encompassing structure.   [GET_GLOBAL_VAR]
     global     => Incfg%get_global_var()
```
This is from a PR from @weiyuan-jiang (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/623) which requires MAPL 2.25.

So the solution is...update to MAPL 2.25. I'll make a PR.

There have been other updates in the GEOSgcm as well, but I'll do those in a separate PR so that @biljanaorescanin can test more fully.

mathomp4:
Closes #577 

This PR updates MAPL to v2.25.0. This is needed to support GEOSgcm_GridComp `develop` which pulled in a PR needing a newer MAPL (https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/623)

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -29,7 +29,7 @@ GMAO_Shared:
 MAPL:
   local: ./src/Shared/@MAPL
   remote: ../MAPL.git
-  tag: v2.23.1
+  tag: v2.25.0
   develop: develop
 
 GEOSgcm_GridComp:


mathomp4:
Note: I'm assuming this is zero-diff. I mean, MAPL didn't really do anything to change answers. 

But, for @gmao-rreichle, you might notice a slight change in History output. MAPL 2.25 adds the `regrid_method` to the variable metadata now (see https://github.com/GEOS-ESM/MAPL/pull/1643). It's possible with crazy enough `nccmp` flags, you could see this as "Files are different!" but it's just due to extra metadata.

gmao-rreichle:
@mathomp4, many thanks for putting this together and for the heads-up about the potential for "fake" non-0-diff results.  I'm running the full suite of tests now.  It'll probably be a few hours before they're done.  I'll report back.

weiyuan-jiang (approved):

gmao-rreichle:
@weiyuan-jiang,
My test results do NOT match yours.  We might have a problem.
Unfortunately, my aggglobal/assim test ran out of time, so I had to redo it, which is why I haven't weighed in yet.   Fortunately, this one passed on the 2nd try.
**However, my gnuglobal/assim test failed the comparison!** I have not yet been able to look into the differences, and I'm very confused to see that your test seems to have passed.  The complicating factor here might be that we never tested some of the recent merges for GNU, because the update of the tests to gcc12 was pending when we moved ahead with the LDAS merges last week.  
We will have to look into this in more detail.  Not sure if I can do anything before the weekend, though.
For now, do **NOT** merge anything else into LDAS.

cc: @mathomp4

weiyuan-jiang:
Hmm, it is interesting. I saw my test succeeded, so I merged. Lets see what will happen tonight.

Issue #582, opened 2022-09-10
Comparison error in GNUGLOBAL/assim test 


gmao-rreichle:
The GNUGLOBAL/assim test conducted by @gmao-rreichle for #578 failed the comparison.
The same test conducted by @weiyuan-jiang passed, as did the automated nightly test of 2022-09-09.
The failed test was moved to here:
/gpfsm/dnb31/rreichle/SystemTests/runs/LDAS_GNUGLOBAL_20220909/
The differences are likely roundoff (confirmed by @gmao-rreichle, 10 Sep 2022 12:03h).
Comparison of the GEOSldas log files of the GNUGLOBAL/assim tests by @gmao-rreichle and @weiyuan-jiang suggests that some variables are not initialized by default in gcc12.  OBS_PARAM_NML differs in some fields such as %SPECIES and %ANG.  The fields that differ should not be used and therefore cannot by themselves explain the comparison failure.  But the lack of initialization in these fields suggests that other variables might also not be initialized in gcc12.  OBS_PARAM_NML is initialized to 0 in the standard-optimization Intel run.  It is not clear if they were initialized in gcc11.  
Maybe the lack of initialization explains the seemingly intermittent nature of the comparison failure?  
It is also possible that something is wrong with the failed test itself, but I don't see what that could be.

gmao-rreichle:
I re-ran the GNUGLOBAL/assim test once yesterday, and it passed, as did the automated nightly test.  I suppose we can just wait to see if the failure happens again in the automated nightly tests.

gmao-rreichle:
Another data point: The automated nightly GNUGLOBAL/assim test of 2022-09-12 failed.  A quick inspection suggests that this could be similar to the failure of my test of 2022-09-09.  I'll take a closer look tomorrow.

gmao-rreichle:
I plotted the 2017-10-16, 0z ObsFcstAna output from the automated GNUGLOBAL/assim test of 2022-09-12 vs. the BASELINE.  Differences are consistent with roundoff.  (The contents of this ObsFcstAna file are from the tail end of the test run and the comparison should fail if there are any differences at all.)  
Interestingly, the automated test of 2022-09-12 is zero-diff vs. my own test run of 2022-09-09.  
Could the result be processor-dependent?  Does the automated test prescribe the processor?

weiyuan-jiang:
In our link.j we specify
#SBATCH --nodes=4 --ntasks-per-node=28
[...]

PR #583, opened 2022-09-14
default values for obs_param_nml

weiyuan-jiang:
Assign default values to obs_param_nml. It should not have any effects on the existing program.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
@@ -248,6 +248,7 @@ contains
     logical :: smap_species, smos_species
 
     type(obs_param_type), dimension(N_obs_species_nml) :: obs_param_nml
+    type(obs_param_type) :: obs_param_default
 
     character(len=*), parameter :: Iam = 'read_ens_upd_inputs'
     character(len=400) :: err_msg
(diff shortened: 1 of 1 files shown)


weiyuan-jiang:
@gmao-rreichle , I am testing this PR.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_upd_routines.F90
```diff
@@ -268,7 +269,40 @@ subroutine read_ens_upd_inputs(               &
...
+    obs_param_default%scalename      = 'NO'
+    obs_param_default%flistpath      = 'NO'
+    obs_param_default%flistname      = 'NO'
+    obs_param_default%errstd         = 4.
+    obs_param_default%std_normal_max = 2.5
+    obs_param_default%zeromean       = .true.
+    obs_param_default%coarsen_pert   = .true.
+    obs_param_default%xcorr          = 0.25
+    obs_param_default%ycorr          = 0.25
+    obs_param_default%adapt          = 0
```
gmao-rreichle, line 302: @weiyuan-jiang, thanks for putting this together.  We should initialize to values that very obviously reflect "UNDEF".
That is, all fields that are strings should be set to "UNDEF".
For floating point numbers, we should use MAPL_UNDEF.  (Or should we stick with -9999.0, the typical LDAS no-data-value?)
Not sure what to use for integers.  Does MAPL have a value for undefined integers?  This gets a bit tricky because the no-data-value depends on what kind of integer it is.  Since all integer fields in obs_param_type are declared as "integer" (presumably defaulting to signed int*4), maybe use -9999 as the no-data-value.
For logicals, ".false." is probably best.
Also, the structure is declared in enkf_types with a vector of length N_ang_max for the field %ang.   That is, we need to initialize all 7 elements of %ang. 
I suppose the initialization should be a subroutine or function that lives in enkf_types.F90.
Having said that, I don't think it really matters whether OBS_PARAM_NML is initialized.  I brought this up in #582 simply because in the gcc12 test setup, there is no initialization by default.  This means that some other variable that actually matters for calculations may likewise not be initialized.

weiyuan-jiang: I am wondering to abandon this PR. It is not a necessary one

gmao-rreichle: That's ok with me.  It's probably better to spend the effort on determining the best way to avoid the nightly test failures.

gmao-rreichle:
Closing per recent conversation above.

PR #584, opened 2022-09-22
support for catchcn ensemble run

weiyuan-jiang:
Add CatchCN-specific export variables to EnsGridComp. 
Required to support ensemble simulations with CatchCN in GEOSldas needed by @gmao-jkolassa.

Dependencies:

- Depends on GCM GC PR https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/645, which in turn depends on MAPL v2.27.1 (See merged PR https://github.com/GEOS-ESM/MAPL/pull/1683.)

Prior to merging, change components.yaml accordingly. 

After merging, change the following input files for the LDAS Nightly tests:
```
/discover/nobackup/mathomp4/LDAS_Restarts/NGHTLY_TST_TV4000/HISTORY_L4_SM.rc
/discover/nobackup/mathomp4/LDAS_Restarts/NGHTLY_TST_CS/assim/ldas179b8_C180/run/HISTORY.rc
```
Specifically, replace `VEGDYN0000` with `VEGDYN_e0000`


weiyuan-jiang:
@gmao-jkolassa , if you want to test ensemble run, please use this fresh checkout of the branch .

src/Applications/LDAS_App/process_hist.csh
```diff
@@ -54,6 +54,13 @@ endif
 if($NENS > 1) then
    set GridComp = ENSAVG
    sed -i 's|VEGDYN|'VEGDYN0000'|g' $HISTRC
+   sed -i 's|CATCHCNCLM45|'CATCHCNCLM450000'|g' $HISTRC
```
gmao-rreichle, line 57: @weiyuan-jiang, @gmao-jkolassa, a couple of things here:
1. Why is there not a similar line for CATCHCNCLM4[0]?  Does the CatchCN ens run only work for CatchCNCLM45?
2. When we decided on CatchCN GridComp and restart names ending in numbers (as in catchcn45), we didn't think ahead to ensemble simulation.  It's rather ugly to have restarts with names such as "catchcnclm450018".  I'm afraid it would be rather disruptive to include a symbol to separate the model version from the ensemble id, as in "catchcnclm45_0000", but it may still be better to make the change now.  It's only going to get more disruptive.  An open question would be if the same change should be applied to "catch" restarts.  I'm not sure what to think.

weiyuan-jiang: > @weiyuan-jiang, @gmao-jkolassa, a couple of things here:
> 
> 1. Why is there not a similar line for CATCHCNCLM4[0]?  Does the CatchCN ens run only work for CatchCNCLM45?
> 2. When we decided on CatchCN GridComp and restart names ending in numbers (as in catchcn45), we didn't think ahead to ensemble simulation.  It's rather ugly to have restarts with names such as "catchcnclm450018".  I'm afraid it would be rather disruptive to include a symbol to separate the model version from the ensemble id, as in "catchcnclm45_0000", but it may still be better to make the change now.  It's only going to get more disruptive.  An open question would be if the same change should be applied to "catch" restarts.  I'm not sure what to think.
> 3. For Nens>1, why should all catchcnclm45 exports come from ens member 0000?  (Maybe/probably I'm missing something here.)

Rolf,

1) I think nobody is using cnclm40 now.
2)This is unusual to export a variable from ensemble member. In general, we can get it from ENSAVG grid comp.
3)I think there is only one from ens member 0000, all others are from ENSAVG. Even this one can be moved to ENSAVG

gmao-jkolassa: Good morning,

`I think nobody is using cnclm40 now.`

That is true. However, if it is not too much trouble I think it would make sense to also add the ensemble capability for CNCLM40. Otherwise, I think it could get very confusing in the future as to why different versions of CatchCN have different functionalities.

` I'm afraid it would be rather disruptive to include a symbol to separate the model version from the ensemble id, as in "catchcnclm45_0000", but it may still be better to make the change now. It's only going to get more disruptive.`

I agree that it would make sense to make this change now. @weiyuan-jiang I think Rolf was referring to the restart files that are created by ldas_setup in /input/restart. They currently will be called catchcnclm45XXXX (with XXXX being the ensemble ID) and catchcnclm45_XXXX would be cleaner. But I am not sure how much work that would actually entail.

`I think there is only one from ens member 0000, all others are from ENSAVG. Even this one can be moved to ENSAVG`

Yes, only CNFROOTC is coming from ensemble member 0000. I think it is handled differently because it is a CNCLM45-specific variable (that is, it does not exist in CNCLM40), so I don't know to what extent that impacts our ability to have it come from ENSAVG like the other variables.

gmao-jkolassa:
This is just to confirm that an ensemble experiment using the current version of this branch works without any issues for both CatchCNCLM40 and CatchCNCLM45. Note that I only tested that it runs. Please let me know if you would like a full science-validation of this.

src/Applications/LDAS_App/process_hist.csh
gmao-rreichle, line 57: I think I understand now why there's no matching entry for CATCHCNCLM40.  So that's fine.  But the `sed` replacement here should **not** be such that we end up pulling the variable CNFROOTC from ens member 0000.  This variable must come from ENSAVG for consistency with all the other outputs from `'GridComp'` (which eventually turns into ENSAVG, right?)
The somewhat awkward templating in `GEOSldas_HIST.rc` may be hitting a wall soon...

gmao-rreichle: For my original items 1. and 3., see my comment below https://github.com/GEOS-ESM/GEOSldas/pull/584#discussion_r979076273

For item 2., we still need to understand better what it takes to change the CatchCN restart names for ensemble simulations from catchcnclm4[5]XXXX to catchcnclm4[5]_XXXX.  Having the CatchCN version number blend seamlessly into the ens id is clearly undesirable.  Maybe this can be limited to CatchCN ens simulations and doesn't have to touch Catch at all and CatchCN single-member runs.  Then there should be no impact on existing experiments and restart names.  But that would create an asymmetry between Catch and CatchCN ensemble simulations.  We can discuss more next week.

src/Applications/LDAS_App/ldas_setup
```diff
@@ -113,6 +113,7 @@ class LDASsetup:
         self.ladas_coupling = 0
         self.in_rstfile = None
         self.in_tilefile = 'None' # default string
+        self.ens_id_width = 6  # _exxxx
```
gmao-rreichle, line 116: @weiyuan-jiang: I think it would be better to use ENS_ID_WIDTH=4 as the default, where 4 refers to the number of **numerical** digits in the ensid_string.  That is, add the extra 2 to the field with (for "__e") as needed in the scripts and source code.  Then existing LDAS.rc files would still work. 
(Of course, a better name for ENS_ID_WIDTH would be ENS_ID_DIGITS, but that would not be backward compatible for existing LDAS.rc files...)

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1271,10 +1265,10 @@ subroutine Initialize(gc, import, export, clock, rc)
        nhms = h *10000 + m*100  + s
 
        do ens = 0, NUM_ENSEMBLE-1
-          call get_id_string(id_string, ens + FIRST_ENS_ID, ens_id_width)
+          call get_ensid_string(ensid_string, ens + FIRST_ENS_ID, ens_id_width, NUM_ENSEMBLE ) ! xxxx not _exxxx
```
gmao-rreichle, line 1268: I don't understand the comment here:   `...  ! xxxx not _exxxx`
Doesn't get_ensid_string produce "_eXXXX"?
[Same comment applies to line ~1591 below.]

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1594,9 +1588,9 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
           read(datestamp(10:13),*) nhms
           nhms = nhms*100
           do ens = 0, NUM_ENSEMBLE-1
-             call get_id_string(id_string, ens + FIRST_ENS_ID, ens_id_width)
+             call get_ensid_string(ensid_string, ens + FIRST_ENS_ID, ens_id_width) !xxxx, not _exxxx
```
gmao-rreichle, line 1591: I don't understand the comment here:   `...  ! xxxx not _exxxx`
Doesn't get_ensid_string produce "_eXXXX"?
[Same comment applies to line ~1268 above.]

src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/GEOS_MetforceGridComp.F90
```diff
@@ -677,13 +678,10 @@ subroutine Initialize(gc, import, export, clock, rc)
...
     if (ensemble_forcing .and. NUM_ENSEMBLE > 1) then
-      id_string = ""
-      call MAPL_GetResource ( MAPL, ens_id_width, Label="ENS_ID_WIDTH:",      DEFAULT=0,       RC=STATUS)
+      ! note: comp_name ends in "_eXXXX"; for GEOS ADAS forcing, extract hard-coded 3-digit ens id string
       k = len(trim(comp_name))
-      id_string = comp_name(k-ens_id_width+1:k)
-      k = len(trim(id_string)) 
-      ! hard coded 3 character for forcing
-      call ESMF_CFIOStrTemplate(ens_forcing_path, trim(adjustl(mf%Path)),'GRADS', xid = trim(id_string(k-2:k)), stat=status)
+      ensid_string3 = comp_name(k-2:k)
```
gmao-rreichle, line 683: This would fail if a user specifies only 2 digits for ens id in LDAS.rc.  Should we add check that ENS_ID_WIDTH must be at least 3?  (not counting the extra 2 char for "_e")

src/Applications/LDAS_App/ldas_setup
weiyuan-jiang: good point. I will change it to 4

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
weiyuan-jiang: yes. For backward compatibility, I didn't change the file name template, so it should be right here.

src/Components/GEOSldas_GridComp/Shared/LDAS_ensdrv_Globals.F90
```diff
@@ -141,7 +143,33 @@ elemental   function LDAS_is_nodata(data) result(no_data)
...
+     if (num_ensemble == 1) then
+        ensid_string = ''
+        return
+     endif
+     
+     if (ens_id_width <=0) then
+       call ldas_abort(LDAS_GENERIC_ERROR, 'get_ensid_string', 'ens_id_width should be greater than 1')
+     endif
+
+     write (fmt_str, "(A2,I1,A1,I1,A1)") "(I", ens_id_width,".",ens_id_width,")"
```
gmao-rreichle, line 167: I think this formatting will fail for ens_id_width>9 because of the "I1" format.  I don't think we need to worry about ens_id_width larger than 9, so I think we should just restrict ens_id_width to be less than 10.
While we're at it, I think we should have ens_id_width be a minimum of 3, then the issue with the number of digits for the LADAS ensemble met forcing coupling being hardwired to 3 is also taken care of. 
So maybe add something like the following (maybe here, or perhaps better where we first read NUM_ENSEMBLE and ens_id_width)
```
if (num_ensemble>) then
_ASSERT(ens_id_width>2 .and. ens_id_width<10,"Must use 3<=ens_id_width<=9 for NUM_ENSEMBLE>1")
end if
```
Does this make sense?

weiyuan-jiang:
We have to change MAPL if we don't add 2 to the ens_id_with.  I think the id should include "_exxxx" even we drop of the backward compatibility.  @gmao-rreichle

weiyuan-jiang:
@gmao-rreichle @gmao-jkolassa It should work now. But the nightly test would fail because we have hard coded HISTORY.rc for global assume and globules assume

gmao-rreichle:
> We have to change MAPL if we don't add 2 to the ens_id_with. I think the id should include "_exxxx" even we drop of the backward compatibility. @gmao-rreichle

@weiyuan-jiang, I don't understand how MAPL comes into this.  I agree that we don't want to change MAPL, but maybe there's another way to distinguish between the number of digits and the total string length (ens_id_width) in LDAS.

weiyuan-jiang:
These two assert should ease your concern if the users still use 4 for the width. Actually, the 6 is used in the lads_setup which should not have problem.
](https://github.com/GEOS-ESM/GEOSldas/blob/a045efd37511be89e3763a97cd3c07b7b3443473/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90#L179-L180)

weiyuan-jiang:
The way we extract the template has to change in MAPL if we change is the width.
https://github.com/GEOS-ESM/MAPL/blob/160a8d720f5357d07d419818db7f53d51d471f2b/generic/MAPL_Generic.F90#L1508-L1509

gmao-jkolassa:
I ran two shorts tests with the latest version of this branch. It builds successfully and runs successfully for both CatchCN40 and CatchCN45.

weiyuan-jiang:
When this branch is merged, the two files for nightly test should be changed too:
/discover/nobackup/mathomp4/LDAS_Restarts/NGHTLY_TST_TV4000/HISTORY_L4_SM.rc
/discover/nobackup/mathomp4/LDAS_Restarts/NGHTLY_TST_CS/assim/ldas179b8_C180/run/HISTORY.rc 

replace VEGDYN0000 with VEGDYN_e0000

@mathomp4 , We will let you know when we merge

gmao-rreichle:
@weiyuan-jiang, @gmao-jkolassa, I applied a hopefully final set of minimal edits https://github.com/GEOS-ESM/GEOSldas/pull/584/commits/f1bd407041126d84aadd2ccffc29069aeffc2a86
I don't think we need to retest the new functionality.  
@weiyuan-jiang, did you run the LDAS Nightly tests with the above-mentioned edits to HISTORY.rc? If we are sure the GEOSldas PR is 0-diff for GEOSldas, we can ask @biljanaorescanin to test the matching GCM GC PR https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/645

and then ask Scott to merge the matching GCM GC PR as soon as he can.

gmao-rreichle (approved):

PR #586, opened 2022-09-28
Further clean up and re-organize make_bcs

weiyuan-jiang:
replace LDAS_ EASE_conv.F90 with EASE_conv.F90 from Raster and clean up some calls related to EASE_conv

depends on https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/634


weiyuan-jiang:
It is highly possible this PR would be non-zero diff 1) cell area is calculated 2) the box ur and ll are calculated now

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -2100,14 +2099,15 @@ subroutine write_smapL4SMaup( option, date_time, work_path, exp_id, N_ens,    &
...
 
     ! --------------------------------------------------------------
     !
     ! smapL4SMaup output only works for 9 km EASE grids
 
-    if ( (trim(tile_grid_g%gridtype)/='EASE_M09'  )  .and.                  &
+    if ( (trim(tile_grid_g%gridtype)/='EASEv1_M09'  )  .and.                  &
          (trim(tile_grid_g%gridtype)/='EASEv2_M09')         )               then
-       err_msg = 'out_smapL4SMaup requires tile-space for 9 km EASE[v2] grid'
+       err_msg = 'out_smapL4SMaup requires tile-space for 9 km EASEv1 or EASEv2 grid'
```
gmao-rreichle, line 2110: @weiyuan-jiang : I tried the nightly tests on the latest branch (https://github.com/GEOS-ESM/GEOSldas/pull/586/commits/5bd729d90ea0bca01bcd1665690e584366f8033e).  
The LDAS_GLOBAL/assim test fails with the following error message, which comes from this if block:
"LDAS ERROR (3000) from write_smapL4SMaup: out_smapL4SMaup requires tile-space for 9 km EASEv1 or EASEv2 grid"
Did the value of `tile_grid_g%gridtype` at this point in the test case change?  Could this be "EASEv2_M09" vs just "M09"?  See also line 2232 below.

weiyuan-jiang: The problem is in LDAS.rc with grid name as SMAP-EASEv2-M09. We either change it to SMAP_EASEv2_M09 or change the code here if (index(...., 'M09') == 0) then

gmao-rreichle: For backward compatibility with existing LDAS.rc files, let's use  `if (index(...., 'M09') == 0)`.  It's extremely unlikely that there'll ever be another, non-EASE gridtype that has the string "M09" in the name.

weiyuan-jiang: sure. changed. Hopefully the nighty test will pass this point.

gmao-rreichle: @weiyuan-jiang, I retested the updated branch.  Unfortunately, I'm still getting a runtime error in the LDAS_GLOBAL/assim test.  GEOSldas exits with the following message:
` LDAS ERROR (3000) from scale_obs_Tb_zscore: tile_id not found in scaling parameter file`
This doesn't make any sense to me.  The tile IDs and scaling parameters should be unchanged in the inputs from the test.  Did we touch the tile IDs anywhere in the branch?

gmao-rreichle:
Limited "nightly" test conducted by @weiyuan-jiang on 12-Oct-2022 after commit https://github.com/GEOS-ESM/GEOSldas/pull/586/commits/ef6fd43068696c549a9b4dbfa9215131ac9d52ec
Branch runs successfully again but fails the comparison for all tests that use the EASE tile space, as expected. 
Postponed full verification of non-0-diff changes until after we have resolved the dependency on https://github.com/GEOS-ESM/GEOSgcm_GridComp/pull/634

gmao-rreichle:
@biljanaorescanin : I don't know if the current "develop" branch (prior to merging this PR) will now work with the updated GCM GC develop branch that includes GCM GC PR #634.  It's possible we need the present GEOSldas PR or GEOSldas develop will no longer work.  On the other hand, this PR is not 0-diff and we left the final testing open.  We'll need to retest this PR as soon as possible.

biljanaorescanin:
Test summary attached:
Screen Shot 2022-12-02 at 4 28 11 PM

gmao-rreichle:
@biljanaorescanin, @weiyuan-jiang, I went through Biljana's tests. 
In the tilegrids.bin files, we have the following difference:
BASELINE:   `gridtype: 'EASEv2_M09'`
CURRENT:   `gridtype: 'EASEv2-M09'`
That is, we changed the underscore to a dash in the gridtype label that is archived in the tilegrids file.  I don't remember if this was intentional.  For reference, the tilegrids file from the Version 7 L4_SM product has an underscore in the gridtype label.
I think we intentionally changed the source code so that both strings will work in GEOSldas runs.    
@weiyuan-jiang, do you remember anything about this?  Where exactly does this string come from when it's dumped into the tilegrids bin file? 

The rest of the differences checks out ok:
The HISTORY and rst differences in the "assim" comparisons are within roundoff and as expected.
The differences in the tilecoord.bin files are roundoff in the field "area" and as expected.
In the tilegrids.bin files there are roundoff differences in the ll_lat, ll_lon, ur_lat, ur_lon, dlon, and dlat numbers, as expected.

weiyuan-jiang:
@gmao-rreichle @biljanaorescanin I think we have changed the name from "_Mxx" to "-Mxx"

gmao-rreichle:
> @gmao-rreichle @biljanaorescanin I think we have changed the name from "_Mxx" to "-Mxx" ...but let me double check. I think we make the names consistent "_Mxx"

@weiyuan-jiang, @biljanaorescanin:  I looked into this a bit more and am now quite confident of the following:

1) I think the primary objective was to accommodate both "-Mxx" and "_Mxx" in existing tilegrids files when they are processed in the source code (for backward compatibility).  This worked fine. 

2) We intentionally changed the "gridname" in the tilegrids file to match what is in the bcs *.til file, which is really the source of everything.  I remember we opted to stick with the less desirable and less consistent "-Mxx" string because this was already used in bcs going back to at least NLv3.  

3) Note that the original "tilegrids" file in Matt's input directory:
`/discover/nobackup/mathomp4/LDAS_Restarts/NGHTLY_TST/model/input/restart/m36_global_regression/output/SMAP_EASEv2_M36_GLOBAL/rc_out/m36_global_regression.ldas_tilegrids.bin`
uses "_Mxx".  When Biljana tested the branch, "gridname" was taken from the bcs file via subroutine LDAS_read_til_file() and inserted into the tilegrids file via subroutine LDAS_create_grid_g() as intended, which correctly reproduces the desired "-Mxx" string in the tilegrids file in the branch. This is an intentional change and should fail the comparison with Matt's present BASELINE experiment.  Weiyuan's most recent commit overrides this, so I reverted back to the branch as tested by 
Biljana. 

4) Once this is merged, the most of the LDAS tests will fail the comparison and need to be re-baselined.  I'll go ahead and merge.  With any luck, @mathomp4 can re-baseline tomorrow.

gmao-rreichle (approved):

gmao-rreichle:
@GEOS-ESM/cmake-team : Can you please approve this PR when you get a chance? 
cc: @weiyuan-jiang @biljanaorescanin

PR #587, opened 2022-10-05
Bug fix: Make "lenkf.j -debug" work (again)

weiyuan-jiang:
pass on debug option through command line
When "lenkf.j -debug" was run at the command line, somehow the arguments or environment conflicted with the "source g5_modules" call within lenkf.j.  This PR fixes the problem.

PR #591, opened 2022-10-13
Clarify help message for "lenkf.j -debug"

weiyuan-jiang:
Clarify help message that users see when running "lenkf.j -debug".


gmao-rreichle (commented):
@weiyuan-jiang, thanks for the proposed change.  I don't think it's a good idea, though.  Totalview may not be available on some compute systems.  Also, the one-line command isn't onerous given that the user still has to go through the totalview option windows and select the MPI stack etc.  It's also good to have the message pointing to the Wiki page with instructions.  The message could, of course, be added back, but if lenkf.j launches totalview, the user is almost certainly going to miss the message because they are then presented with a totalview window.

weiyuan-jiang:
> 

I think the wiki is obsolete. The debugger can not be launched that way.

gmao-rreichle:
> I think the wiki is obsolete. The debugger can not be launched that way. I will update wiki

@weiyuan-jiang, @gmao-qliu:
Please do **not** yet edit the Wiki.  I'm rather confused now.  In #587, there is [comment](https://github.com/GEOS-ESM/GEOSldas/pull/587#issuecomment-1275211573) that things worked for @gmao-qliu, which plainly contradicts @weiyuan-jiang's message above.  It's not clear to me what works and what does not. 

For the record, I'm adding an excerpt from an off-github email by @gmao-qliu on 13-Oct-2022:
"_My understanding is that running lenkf.j will start a new shell which is different from the current one. initiating Totalview within lenkf.j will ensure totalview to start from the same shell as "source g5_modules". Weiyuan can correct me if inaccurate._"

So, can we launch totalview manually from the command line after "source g5_modules"?  That would be a minimal addition to the Wiki and would make the code less specific on what is available on Discover.

weiyuan-jiang:
> can we launch totalview manually from the command line after "source g5_modules"? That would be a minimal addition to the Wiki and would make the code less specific on what is available on Discover.

Yes, we can. So you want "lenkf.j -debug" just echo help information?

gmao-qliu:
> > can we launch totalview manually from the command line after "source g5_modules"? That would be a minimal addition to the Wiki and would make the code less specific on what is available on Discover.
> 
> Yes, we can. So you want "lenkf.j -debug" just echo help information?

yes. that'll work. but can we remove the "source g5_module" part if with "-debug"? It outputs "g5_modules: Setting BASEDIR and modules for .." which is confusing.

src/Applications/LDAS_App/lenkf.j.template
```diff
@@ -38,7 +38,9 @@ endif
 unset argv
 setenv argv
 
-source $GEOSBIN/g5_modules
+if ( $debug_flag == 0 ) then
+  source $GEOSBIN/g5_modules
+endif
```
gmao-rreichle, line 43: This is meant to take care of @gmao-qliu's suggestion to avoid "source g5_modules" when in debug mode.  
If this works, then I'm not sure why we would need `unset argv; setenv argv` above that was introduced to avoid g5_modules tripping up when lenkf.j is called with the -debug flag.  Maybe the suggestion to omit "source g5_modules" here won't work?  I'm getting really confused.

weiyuan-jiang: Since '-debug" option only prints out information, I would like to change it to "--help". It prints out and exits before source g5_modules

gmao-rreichle: Having said that, does the preprocessing possibly need "source g5_modules"? @gmao-qliu: Did you verify that we can skip "source g5_modules" when running with -debug?

weiyuan-jiang: > 

Actually, the preprocessing is not necessary for debug. Except users change the directory to scratch and launch totalview

gmao-qliu: > Having said that, does the preprocessing possibly need "source g5_modules"? @gmao-qliu: Did you verify that we can skip "source g5_modules" when running with -debug?

actually "source g5_modules" is needed with -debug because otherwise $BASEDIR is undefined.  I take my suggestion back.

weiyuan-jiang: After users finish ldas_setup and get the interactive nodes, they can go to the "run" directory, source ../build/bin/g5_modules.sh , then load and launch totalview in run directory

gmao-rreichle: @weiyuan-jiang : Are you seriously saying that the first 200-300 lines in lenkf.j are not needed?  If that's the case, why can't we just delete them altogether?

weiyuan-jiang:
We can discard my changes. But we need to add more information to the wiki. 1) launch totalview in scratch directory 2) config the same number of processors in totalvies as that in lenkf.j

gmao-rreichle:
@weiyuan-jiang, I'm still not sure that for debugging we can simply skip the first 200-300 lines of lenkf.j in all cases.  Let's keep everything as is for now and discuss at our next tag-up.

weiyuan-jiang (commented):
After lenkf.j exit, SCRDIR and GEOSBIN are not defined anymore. May be it is ease to say go to scratch directory,
source ../build/bin/g5_modules.sh 
module load tview
totalview

gmao-rreichle:
> After lenkf.j exit, SCRDIR and GEOSBIN are not defined anymore.

But the "echo" commands that use $SCRDIR and $GEOSBIN are within lenkf.j, so shouldn't the user get the correct dir names printed in the help text on the screen?

weiyuan-jiang:
> > After lenkf.j exit, SCRDIR and GEOSBIN are not defined anymore.
> 
> But the "echo" commands that use $SCRDIR and $GEOSBIN are within lenkf.j, so shouldn't the user get the correct dir names printed in the help text on the screen?

Oh, you are right.  The echo should print out right information. Another suggestions: 1) The users should choose the same number of processors as that in lenkf.j  2) Just launch totalview without executable because usually we debug in parallel mode. The users will have to open a new parallel session.

gmao-qliu:
> > > After lenkf.j exit, SCRDIR and GEOSBIN are not defined anymore.
> > 
> > 
> > But the "echo" commands that use $SCRDIR and $GEOSBIN are within lenkf.j, so shouldn't the user get the correct dir names printed in the help text on the screen?
> 
> Oh, you are right. The echo should print out right information. Another suggestions: 1) The users should choose the same number of processors as that in lenkf.j 2) Just launch totalview without executable because usually we debug in parallel mode. The users will have to open a new parallel session.

@weiyuan-jiang @gmao-rreichle's suggestion to start totalview with executable and edit the parallel setting afterwards using Ctrl+A works. The current instruction on the wiki page is overall good to me. Someone who tries to run debugging the first time might have more insight of the clarity. One minor edit suggestion: add the highlighted word to the line "To use debugging tools at NCCS, you must be on an interactive compute node, which can be obtained with the following **sample** command:"

gmao-rreichle (approved):
FYI, I further edited the Wiki page to reflect the latest changes by @weiyuan-jiang and @gmao-qliu.

Issue #607, opened 2022-12-07
ulimit settings required for ldas_setup on Discover


amfox37:
As a new user on Discover I was having a somewhat cryptic segfault when running ldas_setup as this is done on a login node and I didn't have:
```
ulimit -s unlimited
ulimit -v unlimited
```
in my .bashrc
Was wondering if we should somehow incorporate these settings into ldas_setup, or at least add a warning about this to the [readme](https://github.com/GEOS-ESM/GEOSldas#readme)? 
Or do we just expect folks to already have this in their .bashrc? 
What do you think @weiyuan-jiang ?

weiyuan-jiang:
@mathomp4  Where do you think it is a better place for this?  g5_modules or rc setting files?

mathomp4:
@weiyuan-jiang We usually put this in the script being run itself. For GEOSgcm, `gcm_setup` doesn't really do anything (pure csh), but in `gcm_run.j` we have `limit stacksize unlimited` (as it's a csh script as well).

The fun is that `ldas_setup` is python now, right? But it is calling commands with subprocess. If what it is calling is itself a shell script, then you'd want to put the right `limit` or `ulimit` calls in those scripts.

If you are calling an executable directly, you might have to do something like:

```python
import resource
resource.setrlimit(resource.RLIMIT_STACK, (resource.RLIM_INFINITY, resource.RLIM_INFINITY))
```

I think this is the [Pythonic equivalent](https://docs.python.org/3.9/library/resource.html) if `ulimit -s unlimited`. I'm not sure what the virtual memory one is. Maybe `resource.RLIMIT_VMEM` or `resource.RLIMIT_AS`?

@amfox37 Do you know what command `ldas_setup` was trying to run when you got the error?

amfox37:
My understanding is I was getting a seg fault when preprocess_ldas.x  called with c_f2g on line 726 enters read_cat_param

mathomp4:
@amfox37 Okay. Makes sense. Perhaps if the `setrlimit` command was done at the top of `ldas_setup` that might be enough? I don't think it needs to be done before every call.

gmao-rreichle:
Addressed with #620

PR #620, opened 2023-01-18
set unlimited stack size in Python script

weiyuan-jiang:
This should solve issue #607


gmao-rreichle:
Thanks, @weiyuan-jiang, for putting together the PR.
@amfox37: When you get a chance, please try again with this fix (after removing the "ulimit" commands from your bashrc) and report back.

amfox37:
Yes, seems good. I was just able to reproduce the previous error with main, then switched to this branch and then set up and run of a test case went smoothly. Thanks @weiyuan-jiang

PR #623, opened 2023-01-23
Add GEOS_Util repo

mathomp4:
With the split of GEOS_Util from GMAO_Shared, the develop branch is now not happy since it points to the `main` branch of GMAO_Shared and now the ldas doesn't see any of GEOS_Util.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -26,6 +26,12 @@ GMAO_Shared:
   sparse: ./config/GMAO_Shared.sparse
   develop: main
 
+GEOS_Util:
+  local: ./src/Shared/@GMAO_Shared/@GEOS_Util
+  remote: ../GEOS_Util.git
+  tag: v1.0.1
+  develop: main
+
 MAPL:
   local: ./src/Shared/@MAPL
   remote: ../MAPL.git


gmao-rreichle:
@mathomp4, thanks for getting the PR underway.   I suppose adding the new GEOS_Util repo to GEOSldas would be the quickest fix.  But I'm not sure it's what we want. 

If we're reorganizing GMAO_Shared and GEOS_Util, I'm not sure "remap_restart" should remain in GEOS_Util/**post**/remap_restart/.  This utility can be thought of as post-processing, but to me it's equally sensible as a pre-processing utility.
Would it make more sense to add the restart remapping code to the GCM GridComp [Utils/mk_restarts](https://github.com/GEOS-ESM/GEOSgcm_GridComp/tree/develop/GEOSagcm_GridComp/GEOSphysics_GridComp/GEOSsurface_GridComp/Utils/mk_restarts) directory?  Then the restart-related code would be in one place (for the most part).  The downside may be that it might be harder to phase out the old "regridding" tools if the new tools aren't sitting right next to it.

@sdrabenh: What do you think?

gmao-rreichle:
From @sdrabenh:

> We had a couple reasons for splitting off GEOS_Util. First, Ricardo always preferred stand-alone GMAO_Shared releases to isolatie plot updates – I guess so he could cherry-pick things more easily. Second, most things outside of GEOS_Util are more relevant to the DAS and it didn’t make sense for me to be deciding the fate of those pull requests. Ricardo had long suggested breaking this out into a separate repository. I tend to agree with you that these regridding tools are really more preprocessing than postprocessing, so I would think the proper place for them would be in @GEOS_Util/pre. Of course that begs the question should mk_restarts also be there as more of a stand-alone package? As far as the perl and python tools remaining next to each other, they still would in “./install/bin” which is the proper directory to be using them. We could even move the pearl script into pre so they would sit together.

It makes sense to me to break up GMAO_Shared.  The two elements of interest to GEOSldas are:

1) remap_restarts: Since the underlying Fortran code is very closely tied to the model code (at least for Catchment), I don't think we want to move remap_restarts out of the GCM GC.  I didn't check what would be involved in making the underlying utilities work outside of the GCM GC, but my gut feeling is that it would require some duplication of model stuff in GEOS_Util.  If someone can prove me wrong, I'm happy to have all of the restart code in GEOS_Util.  Failing that, I'd add the remap_restarts python scripts into the GCM GC. 

2) Does the DAS use the surface layer science code in GMAO_Shared/GEOS_Shared?  My (uneducated) guess is that this is model code more than DAS code.  This code is the reason that GEOSldas taps into the development ("main") branch of GMAO_Shared.   Arguably, this code hasn't changed in ages, but that doesn't mean it won't ever change.  Is there a better place for the code in "GEOS_Shared" than in the GMAO_Shared repo?   

cc: @biljanaorescanin @weiyuan-jiang @mathomp4

gmao-rreichle:
> I would advocate for tools to regrid restarts be located in @GEOS_Util/pre since they are more than just surface (upper air) and really do not belong in GEOSsurface_GridComp/Utils.

Good point.  I guess I was outing myself as surface-centric ;)    Let's move remap_restarts python scripts to GEOS_Util/pre then.
I'm still hesitant to think that the underlying Fortran could be moved to GEOS_Util without drawbacks.

Any thoughts on the science code in GEOS_Shared?

sdrabenh:
> > I would advocate for tools to regrid restarts be located in @GEOS_Util/pre since they are more than just surface (upper air) and really do not belong in GEOSsurface_GridComp/Utils.
> 
> Good point. I guess I was outing myself as surface-centric ;) Let's move remap_restarts python scripts to GEOS_Util/pre then. I'm still hesitant to think that the underlying Fortran could be moved to GEOS_Util without drawbacks.
> 
I'm not too worried since those tools are designed to work on restarts dating back to MERRA, so it's not really tied to a specific version of the `GEOSgcm_GridComp` per se. For future restart changes, the user will have to grab a more recent release of `GEOS_Util` anyway. Typically, users want the most recent release anyway.

> Any thoughts on the science code in GEOS_Shared?

Good question. Scanning through `GEOS_Shared` most files haven't been touched since we moved to git. I don't think DAS is doing much with these files based on a recent tag by @rtodling :
https://github.com/GEOS-ESM/GMAO_Shared/compare/v1.6.5...rt_g530_1.6.0_merge
Most changes are outside of `GMAO_Shared`

tclune (approved):

gmao-rreichle:
> Scanning through GEOS_Shared most files haven't been touched since we moved to git. I don't think DAS is doing much with these files

If the GEOS_Shared files are used by the model but not the DAS, and if the point of the new repo is to have GMAO_Shared be more focused on DAS code, then wouldn't it make sense to move GEOS_Shared into GEOS_Util, the GCM GridComp, or some other, more model-specific repo?  

GEOSldas uses "GEOS_Shared/surface_layer.F90" (which must also be used by the GCM), which is pure model code.  
Many of the other routines in GEOS_Shared seem to just support surfacelayer.F90, including fairly basics physics (e.g., "qsat.code"), although there are also other bits with which I'm not familiar.

@mathomp4, @sdrabenh:  Would it be feasible to also remove GEOS_Shared from GMAO_Shared while we're at it?  I don't know where it should go, but it doesn't seem to fit into GMAO_Shared.  If we can move GEOS_Shared, we could remove the GEOSldas dependency on the development ("main") branch of GMAO_Shared.  Please advise, thanks.

mathomp4:
@gmao-rreichle As long as the higher-ups are happy, I can pull GEOS_Shared into its own repo as well. Of course, then you get fun things like two repos named `GMAO_Shared` and `GEOS_Shared`. But I suppose now we have `GMAO_Shared/GEOS_Shared` anyway so... 🤷🏼 

But if @wmputman, @rtodling, @sdrabenh, etc. agree on it then I can pull it out as well.

sdrabenh:
@gmao-rreichle I'm not sure why `GEOS_Shared` should be moved at all, especially if there is not a clear place where it belongs that is better. It is shared by definition of its own name, and I'm sure these routines are not strictly limited to the gcm. I don't necessarily agree that `GMAO_Shared` is going to be DAS-centric as much as `GMAO_Util` was quite unique for the modeling group and contained many of @lltakacs scripts including the plot package that we maintain which justified the succession. Additionally, I'd be very hesitant to add yet another sub-repo to a very long list now, especially given how confusing the names would be between the existing `GMAO_Shared` and a proposed `GEOS_Shared`. If the criteria to remove the dependency of the main branch for `GEOSldas` is `GEOS_Util`, why can't we safely do that now? Aside from 2 files changes 2+ years ago, there haven't been any changes in this directory since we switched to git over 3 years ago. Therefore, it may not be necessary to stay on the bleeding edge of development if it is only this directory. If `GEOSldas` could be moved off the main branch in lieu of a tag, there wouldn't be so many unforeseen issues with changes. @mathomp4 is it possible the specific files here would require @gmao-rreichle approval for changes? Regardless of that, I would be in favor of leaving `GEOS_Shared` where it is now unless there is a problem.

gmao-rreichle:
> I can pull GEOS_Shared into its own repo 

I wasn't really thinking a new repo.  Rather, maybe there is a place within an existing repo where this code might live?  To me, this stuff looks like model code that could fit somewhere in GCM GC.

gmao-rreichle:
Sorry, I had composed and sent off my previous comment before seeing @sdrabenh's comment..

> If the criteria to remove the dependency of the main branch for GEOSldas is GEOS_Shared, why can't we safely do that now?

The dependency exists because GEOS_Shared/surfacelayer.F90 (surface turbulence) is an integral part of the model.  The LDAS needs to stay on the bleeding edge of anything to do with the model's surface, incl. surfacelayer.F90.   The fact that surfacelayer.F90 hasn't been touched in ages doesn't guarantee it's set in stone.  If surfacelayer.F90 remains where it is now, GEOSldas will have to keep pointing to GMAO_Shared "main".  That's not the end of the world.  I just thought it would make sense to check if the modeling group really wants surfacelayer.F90 to remain in GMAO_Shared.

sdrabenh:
> The dependency exists because GEOS_Shared/surfacelayer.F90 (surface turbulence) is an integral part of the model. The LDAS needs to stay on the bleeding edge of anything to do with the model's surface, incl. surfacelayer.F90. The fact that surfacelayer.F90 hasn't been touched in ages doesn't guarantee it's set in stone. If surfacelayer.F90 remains where it is now, GEOSldas will have to keep pointing to GMAO_Shared "main". That's not the end of the world. I just thought it would make sense to check if the modeling group really wants surfacelayer.F90 to remain in GMAO_Shared.

@gmao-rreichle I agree, just because it hasn't changes in while does not guarantee it won't. If @mathomp4 says it's possible to require your approval for PRs touching [surfacelayer.F90](https://github.com/GEOS-ESM/GMAO_Shared/blob/main/GEOS_Shared/surfacelayer.F90) or even [GEOS_Shared](https://github.com/GEOS-ESM/GMAO_Shared/tree/main/GEOS_Shared), would that work for the land group? I would really like to get you all off the main branch so you don't get hit over the head with changes quite as often as you do now.

gmao-rreichle:
> add an entry in [.github/CODEOWNERS](https://github.com/GEOS-ESM/GMAO_Shared/blob/main/.github/CODEOWNERS) for [surfacelayer.F90](https://github.com/GEOS-ESM/GMAO_Shared/blob/main/GEOS_Shared/surfacelayer.F90) so that any changes would have to be approved by land team

That's an option, of course, although the approval requirement would need to apply to all of the files that are used by surfacelayer.F90 (or all of GEOS_Shared for simplicity).  
The benefit of this approach is that the land team would be alerted to any changes in GEOS_Shared *before* they are merged into "main".  The downside is that it would require manual intervention (i.e., testing) each time GEOS_Shared is touched.  
With the dependency on the "main" branch, on the other hand, the testing of GMAO_Shared "main" in GEOSldas happens automatically with the LDAS Nightly tests, albeit after the fact. 

>  I would really like to get you all off the main branch so you don't get hit over the head with changes quite as often as you do now.

Since there have no changes to GEOS_Shared, this wasn't really an issue, and the changes of last Friday were detected by the LDAS Nightly tests as designed.  Most of what's happening in GMAO_Shared is actually of no consequence for GEOSldas, and that is the case in both of the above options.

To summarize: Either way is fine with me.  All we need is a formal and automated process that alerts the land team of any changes in GEOS_Shared.  I'll leave it up to @mathomp4 and @sdrabenh to pick which option works better as a whole.

mathomp4:
@gmao-rreichle Well, `surfacelayer.F90` seems to depend on MAPL and `DragCoefficients.F90`.

I could add any/all of:
* surface-preproc-team
* land-team
* landice-team
* etc.

as an (additional) CODEOWNER to those files if you like. I need to update the GMAO_Shared `CODEOWNERS` anyway, so let me know what you'd prefer.

There is also `atmOceanIntLayer.F90` which is...so many teams? Land? Atmosphere? Ocean? I mean, at least @sanAkel ...

mathomp4:
Honestly, maybe letting the Land Team know anytime *anything* is touched in GMAO_Shared is a good thing. I mean, good ol `qsat` seems to affect everyone!

gmao-rreichle:
@mathomp4:
Since `surfacelayer.F90` is not preprocessing, it should be "land-team".  The dependency extends beyond `DragCoefficients.F90`.  See `#include` statements within the latter.   

> letting the Land Team know anytime anything is touched in GMAO_Shared is a good thing. I mean, good ol qsat seems to affect everyone!

Yes, but I don't think we need to know anything in GMAO_Shared.  Anything in **GEOS**_Shared should be sufficient.

sdrabenh:
I think a key difference here is that `GEOSldas` is an operational system with dependencies in `GEOS_Shared`. I don't think we need approvals from a whole bunch of teams or the entire shared repo, just the minimum necessary to catch anything before it impacts products. Nightlies and CI checks should be sufficient for everyone else. Let's add the `land-team` as a code owner of `GEOS_Shared` and we can finally have `GEOSldas` using a tag. Yay!

gmao-rreichle:
I removed the dependency on the GMAO_Shared "main" branch: https://github.com/GEOS-ESM/GEOSldas/pull/623/commits/9cba134509d6ec896e3397fb99efe4cc315c6579

@mathomp4: Please go ahead and add the land-team as CODEOWNERS of GMAO_Shared/GEOS_Shared

@biljanaorescanin:  Please test this PR.  If all tests pass, please approve and merge. 

@sdrabenh: If you still think it's a good idea, please move GEOS_Util/**post**/remap_restart to GEOS_Util/**pre**.  Since the present PR points GEOSldas to the existing GEOS_Util v1.0.1 release, we can worry later about any changes that this might require for GEOSldas (likely not an issue anyway since we should only need executables from GEOS_Util). 

I hope this resolves everything.  Please let me know if you have any questions or concerns.  I wouldn't be surprised if I got something wrong again...

sdrabenh:
> @sdrabenh: If you still think it's a good idea, please move GEOS_Util/**post**/remap_restart to GEOS_Util/**pre**. Since the present PR points GEOSldas to the existing GEOS_Util v1.0.1 release, we can worry later about any changes that this might require for GEOSldas (likely not an issue anyway since we should only need executables from GEOS_Util).

I'm fine with moving all restart regridding tools into `@GEOS_Util/pre` since that is more logical. That think that should be rather straightforward since the files are in the same repo.

sanAkel:
Good catch @mathomp4 'atmOceanIntLayer.F90' can go wherever-- new place. In a few years, I will deprecate its functionality.

biljanaorescanin:
All tests passed.

biljanaorescanin (approved):

PR #628, opened 2023-03-01
 heterogenous standard deviation for perturbation input from  file

saraqzhang:
(1) code fixed and tested  for the option of input perturbation sdv. 
(2) input file structure explained in  notes inline.  

note: Currently users need to read the src code to know the expected file structure. it may be helpful  if  a sample  netcdf file and a reading/writing f90 code  be added somewhere  like in Util or App.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
@@ -658,7 +658,9 @@ contains
     else ! cubed-sphere grid
        !for cubed-sphere grid, global lat_lon grid
        N_x=tile_grid%n_lon
-
+       ! NOTE: the grid specification is hard-wired here, 
+       ! if perturbation sdv is heterogeous input from a file  
+       ! the input grid must match this grid  (sqz 2/2023)  
        n_lon=4*N_x
        n_lat=3*N_x
        write(lattmp,'(I6.6)') n_lat
@@ -1111,6 +1113,8 @@ contains
        call MPI_BCAST(stdfilename_force_pert,300,MPI_CHARACTER,0,mpicomm,mpierr)
 
        nc4_file = stdfilename_force_pert
+       ! NOTE: the input file is in netcdf, with a group 'std_force_pert',
+       ! and the grid is same as *global* grid (tile_grid_g) (sqz 2/2023) 
 
        ! --compute-local-shape-first-
        ! ASSUMPTION: data in file are on the *global* grid (tile_grid_g)
@@ -1150,11 +1154,11 @@ contains
                         )
                    if (nc4_stat /= nf90_noerr) call handle_nc4_stat(nc4_stat)
                    ! get _FillValue for nc4_varname
-                   nc4_stat = nf90_get_att(nc4_grpid, nc4_varid, '_FillValue', nc4_fillval)
-                   if (nc4_stat /= nf90_noerr) call handle_nc4_stat(nc4_stat)
+                   !nc4_stat = nf90_get_att(nc4_grpid, nc4_varid, '_FillValue', nc4_fillval)
+                   !if (nc4_stat /= nf90_noerr) call handle_nc4_stat(nc4_stat)
                    ! replace _FillValue by zero
-                   where (abs(std_force_pert(ivar,:,:)-nc4_fillval)<nodata_tolfrac_generic) &
-                        std_force_pert(ivar,:,:) = 0.
+                   !where (abs(std_force_pert(ivar,:,:)-nc4_fillval)<nodata_tolfrac_generic) &
+                   !    std_force_pert(ivar,:,:) = 0.
                 end if
              end do
           end if
@@ -1503,6 +1507,8 @@ contains
        call MPI_BCAST(stdfilename_progn_pert,300,MPI_CHARACTER,0,mpicomm,mpierr)
 
        nc4_file = stdfilename_progn_pert
+       ! NOTE: the input file is in netcdf, with a group 'std_progn_pert',
+       ! and the grid is same as *global* grid (tile_grid_g) (sqz 2/2023)  
 
        ! --compute-local-shape-first-
        ! ASSUMPTION: data in file are on the *global* grid (tile_grid_g)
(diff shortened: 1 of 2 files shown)


src/Components/GEOSldas_GridComp/Shared/LDAS_TileCoordRoutines.F90
```diff
@@ -427,7 +427,7 @@ subroutine LDAS_create_grid_g( gridname, n_lon, n_lat,       &
        
        if (date_line_on_center) then
           
-          tile_grid%ll_lon = -180. - tile_grid%dlon/2.
+          tile_grid%ll_lon = -180. + tile_grid%dlon/2.  ! fixed sign (sqz 2023)
```
gmao-rreichle, line 430: @saraqzhang,  
I don't understand this change.  The original (minus) sign seems correct to me. 
The key here is that "ll" is the "lower left" corner of the grid cell and "ur" is the upper right corner of the grid cell.  The westernmost grid cell's western boundary should be at ll_lon=-180-dlon/2.
For, say, dlon=0.5 deg, the grid cell extends from -180.25 to -179.75, with -180.25 being the longitude of the lower-left corner.  
Has this branch/PR been zero-diff tested? 
cc: @weiyuan-jiang, @biljanaorescanin

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
```diff
@@ -1150,11 +1154,11 @@ subroutine get_force_pert_inputs( pert_grid_l,                         &
...
                    ! get _FillValue for nc4_varname
-                   nc4_stat = nf90_get_att(nc4_grpid, nc4_varid, '_FillValue', nc4_fillval)
-                   if (nc4_stat /= nf90_noerr) call handle_nc4_stat(nc4_stat)
+                   !nc4_stat = nf90_get_att(nc4_grpid, nc4_varid, '_FillValue', nc4_fillval)
+                   !if (nc4_stat /= nf90_noerr) call handle_nc4_stat(nc4_stat)
                    ! replace _FillValue by zero
-                   where (abs(std_force_pert(ivar,:,:)-nc4_fillval)<nodata_tolfrac_generic) &
-                        std_force_pert(ivar,:,:) = 0.
+                   !where (abs(std_force_pert(ivar,:,:)-nc4_fillval)<nodata_tolfrac_generic) &
+                   !    std_force_pert(ivar,:,:) = 0.
```
gmao-rreichle, line 1161: @saraqzhang, 
Why comment out these lines?  Without them, the input file will have to have physically sensible values of, say, std_progn_pert for all grid cells, including ocean.   I suppose we could construct input files that have 0 over ocean.  If we do that, then these lines don't do anything.  But if the std_progn_pert inputs contain 1.e15 or -9999, chances are that subsequent calculations using very large or negative numbers might crash.  The lines in question here should prevent this from happening.

src/Components/GEOSldas_GridComp/Shared/LDAS_TileCoordRoutines.F90
saraqzhang: @gmao-rreichle ,   in this case I misunderstood the "ll" , it meant to extend beyond -180.   This part of the code (DC ) is not used in "zero diff" test I ran with cubed sphere grid, which used DE ,PE , not PC, DC.

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
saraqzhang: @gmao-rreichle ,  the current file is created with  the default  sdv value as the base  ( the numbers in ensprop.nml ) everywhere,  then adding modification to certain region. I have not included land mask information in the python code I  made, therefore no  attribute fill value set in the file.  We certainly can add  fillval to attributes when creating the file. and keep those fillval lines the code here.

saraqzhang: @gmao-rreichle ,  the fill value attributes are added to the sdv nc4 file.  The lines reading  fill values are reinstalled and tested with the new nc4 file input.   the comment "!' on those lines can be removed.  I pushed the changes to this branch, but don't know how this PR  be updated.

gmao-rreichle: @saraqzhang, thanks, no need for further updates.  By pushing the branch to github, the PR automatically updates.  Since the only difference (other than added comments) is the location of the nc4 close statement that doesn't get used in any of the tests, the PR should be trivially 0-diff.  I'll work with @biljanaorescanin on when/how we can best merge it.  It may be best to bundle it with another branch/PR that will be tested for 0-diff, just to be safe.

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
```diff
@@ -1111,6 +1113,8 @@ subroutine get_force_pert_inputs( pert_grid_l,                         &
        call MPI_BCAST(stdfilename_force_pert,300,MPI_CHARACTER,0,mpicomm,mpierr)
 
        nc4_file = stdfilename_force_pert
+       ! NOTE: the input file is in netcdf, with a group 'std_force_pert',
+       ! and the grid is same as the *global* grid (tile_grid_g) (sqz 2/2023)
```
gmao-rreichle, line 1117: @saraqzhang, I'm a bit confused by this statement.  In line 661 above, there's a comment that when the model is on the cube-sphere tile space, the pert grid is a hard-wired lat/lon grid (with approximately the same resolution).  The comment here (lines 1116-1117) and again below (lines 1510-1511) suggest that the pert grid is always tile_grid_g.  But for the cube-sphere tile space, this isn't correct.  Maybe I'm misunderstanding this?

weiyuan-jiang: The tile_grid_g here is already a lat-lon grid generated here  https://github.com/GEOS-ESM/GEOSldas/blob/eef59e2d1b69ce8a91122cbbb4a641bdecc810f6/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90#L632

saraqzhang: this comment should be more clear:   the grid in netcdf file is lat-lon grid, the resolution should match what "get_pert_grid(tile_grid_g, latlon_tmp_g) "   generated.   
it is left to user to find out what the resolution should be.  ( for instance in a C180 run, it is 720x540)

gmao-rreichle (approved):
re-approving after edits of comments

PR #634, opened 2023-03-08
update legacy bcs path

gmao-rreichle:
This PR updates the path to the legacy bcs data by pointing to the new "bcs_shared" directory in the GMAO project space on Discover (replacing /discover/nobackup/ltakacs/bcs/).

This is the first step in using (legacy) bcs from the new "bcs_shared" dir.  The next and much more complex step will address using bcs in their revised sub-directory structure.

### DIFF
--- a/doc/README.MetForcing_and_BCS.md
+++ b/doc/README.MetForcing_and_BCS.md
@@ -329,9 +329,9 @@ COMMONLY USED boundary conditions (bcs):
 
 #### MERRA2
 ```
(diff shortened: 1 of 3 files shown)


biljanaorescanin:
@gmao-rreichle  To me these look right. These are only ones I found as well.

gmao-rreichle:
> @gmao-rreichle To me these look right. These are only ones I found as well.

Ok, thanks.  I suggest you go ahead and approve/merge the PR.

biljanaorescanin (approved):

PR #637, opened 2023-03-10
clean up variable and subroutine names for Ldas grid 

weiyuan-jiang:
Use pert_grid as the standard names for perturbation and assimilation

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
@@ -636,8 +636,6 @@ contains
           pert_grid_f = get_minExtent_grid(N_catf, tile_coord_f%hash_i_indg, tile_coord_f%hash_j_indg, &
                tile_coord_f%min_lon, tile_coord_f%min_lat, tile_coord_f%max_lon, tile_coord_f%max_lat, &
                pert_grid_g)
-            
-       else
        endif
     endif
     
--- a/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
@@ -632,11 +632,6 @@ contains
        !  cube-sphere               !  lat/lon grid of resolution similar to that of (cube-sphere) tile_grid
        
        pert_grid_g = get_pert_grid(tile_grid_g)
-       ! pert_grid_f = get_pert_grid(tile_grid_f) -- SEEMS REDUNDANT, OVERWRITTEN BELOW: pert_grid_f = get_minExtent_grid()
-       
-       !if(trim(grid_type) == "Cubed-Sphere" ) then
-       !
-       !   _ASSERT(index(tile_grid_g%gridtype, 'c3') /=0, "tile_grid_g does not describe a cubed-sphere grid")
        
        if ( .not. (pert_grid_g==tile_grid_g) ) then
 
@@ -668,10 +663,7 @@ contains
     call MPI_BCAST(tile_coord_f,N_catf,    MPI_tile_coord_type,0,mpicomm, mpierr)
     call MPI_BCAST(pert_grid_g, 1,         MPI_grid_def_type,  0,mpicomm, mpierr)
     call MPI_BCAST(pert_grid_f, 1,         MPI_grid_def_type,  0,mpicomm, mpierr)
-    
-    !if(trim(grid_type) == "Cubed-Sphere" ) then
     call MPI_BCAST(tile_grid_g, 1,         MPI_grid_def_type,  0,mpicomm, mpierr)
-    !endif
 
     block
       integer, allocatable :: f2tile_id(:), tile_id2f(:)
@@ -711,11 +703,7 @@ contains
     tcinternal%pgrid_g = pert_grid_g
     tcinternal%pgrid_f = pert_grid_f
     tcinternal%pgrid_l = pert_grid_l
-    !if(trim(grid_type) == "Cubed-Sphere" ) then
     tcinternal%tgrid_g = tile_grid_g
-    !else
-    !  tcinternal%grid_g = tcinternal%pgrid_g
-    !endif
 
     do i = 1, NUM_ENSEMBLE
        call MAPL_GetObjectFromGC(gcs(METFORCE(i)), CHILD_MAPL, rc=status)
--- a/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
@@ -620,22 +620,43 @@ contains
        close(10)
        call io_grid_def_type('w', logunit, tile_grid_f, 'tile_grid_f')
 
+       ! get a grid for perturbations and EnKF:
+       !
+       !  tile grid                 !  pert grid
+       !  (defines tile space)      !  (used for perturbations and as "hash" grid in EnKF analysis)
+       ! ===========================================================================================================
+       !  lat/lon                   !  same as tile_grid (i.e., lat/lon)
+       ! -----------------------------------------------------------------------------------------------------------
+       !  EASEv[X]                  !  same as tile_grid (i.e., EASE)
+       ! -----------------------------------------------------------------------------------------------------------
+       !  cube-sphere               !  lat/lon grid of resolution similar to that of (cube-sphere) tile_grid
+       
        pert_grid_g = get_pert_grid(tile_grid_g)
-       pert_grid_f = get_pert_grid(tile_grid_f)
+       ! pert_grid_f = get_pert_grid(tile_grid_f) -- SEEMS REDUNDANT, OVERWRITTEN BELOW: pert_grid_f = get_minExtent_grid()
+       
+       !if(trim(grid_type) == "Cubed-Sphere" ) then
+       !
+       !   _ASSERT(index(tile_grid_g%gridtype, 'c3') /=0, "tile_grid_g does not describe a cubed-sphere grid")
+       
+       if ( .not. (pert_grid_g==tile_grid_g) ) then
 
-       if(trim(grid_type) == "Cubed-Sphere" ) then
-
-          _ASSERT(index(tile_grid_g%gridtype, 'c3') /=0, "tile_grid_g does not describe a cubed-sphere grid")
-         
-          !1) get hash index for cubed-sphere
+          ! arrive here when tile_grid_g is cube-sphere and pert_grid_g is lat/lon after call to get_pert_grid() above
+ 
+          !1) get pert_i_indg, pert_j_indg for tiles in (full) domain relative to pert_grid_g
           do i = 1, N_catf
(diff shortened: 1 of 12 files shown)


src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
```diff
@@ -1955,7 +1955,7 @@ end subroutine output_smapL4SMaup
   ! **********************************************************************
 
   subroutine write_smapL4SMaup( option, date_time, work_path, exp_id, N_ens,    &
-       N_catl, N_catf, N_obsl, tile_coord_f, tile_grid_g, N_catl_vec, low_ind,  &
+       N_catl, N_catf, N_obsl, tile_coord_f, pert_grid_g, N_catl_vec, low_ind,  &
```
gmao-rreichle, line 1958: This subroutines is custom-designed for the SMAP L4_SM product and works **only** if the model is on the EASEv[*] M09 tile space.  It maps data to/from the M09 grid for the purpose of HISTORY-like output.  That is, here we really should use "tile_grid_g", not "pert_grid_g", I think, and make sure that "tile_grid_g" is passed down all the way from the top.   Since thus far we used pert_grid=tile_grid when the model was running in the EASE tile space, this change will not impact the calculations.  But it would be cleaner to use "tile_grid_g" here.  Does this make sense?

gmao-rreichle:
Thanks, @weiyuan-jiang, for putting this together.  I applied a couple of small commits that should further clean up the code and hope I didn't break anything.  I'm not sure the renaming is done in all the right places quite yet.  See inline comment above.  Additionally, perhaps we need to rename tile_grid -> pert_grid in subroutine get_obs_pred().   I haven't really finished this review and will need to revisit the PR later, but I wanted to write down my thoughts up to this point.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_enkf_update.F90
gmao-rreichle: >  all perturbations and assimilations are on pert_grid

Yes.  But this particular subroutine (`write_smapL4SMaup`) does output on the model grid for the specific case of L4_SM on the M09 EASE grid.  It's an odd case.

src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
```diff
@@ -620,22 +620,43 @@ subroutine Initialize(gc, import, export, clock, rc)
...
-       pert_grid_f = get_pert_grid(tile_grid_f)
-
-       if(trim(grid_type) == "Cubed-Sphere" ) then
+       ! pert_grid_f = get_pert_grid(tile_grid_f) -- SEEMS REDUNDANT, OVERWRITTEN BELOW: pert_grid_f = get_minExtent_grid()
+       
+       !if(trim(grid_type) == "Cubed-Sphere" ) then
+       !
+       !   _ASSERT(index(tile_grid_g%gridtype, 'c3') /=0, "tile_grid_g does not describe a cubed-sphere grid")
+       
+       if ( .not. (pert_grid_g==tile_grid_g) ) then
```
gmao-rreichle, line 641: @weiyuan-jiang: Note the change in the if condition here.  The tasks within the if block need to get done whenever pert_grid differs from tile_grid.  I don't think we should make this conditional on the CubeSphere grid.  It just happens that pert_grid differs from tile_grid when we're in CubeSphere tile space.

src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
```diff
@@ -620,22 +620,43 @@ subroutine Initialize(gc, import, export, clock, rc)
...
+       ! -----------------------------------------------------------------------------------------------------------
+       !  EASEv[X]                  !  same as tile_grid (i.e., EASE)
+       ! -----------------------------------------------------------------------------------------------------------
+       !  cube-sphere               !  lat/lon grid of resolution similar to that of (cube-sphere) tile_grid
+       
        pert_grid_g = get_pert_grid(tile_grid_g)
-       pert_grid_f = get_pert_grid(tile_grid_f)
-
-       if(trim(grid_type) == "Cubed-Sphere" ) then
+       ! pert_grid_f = get_pert_grid(tile_grid_f) -- SEEMS REDUNDANT, OVERWRITTEN BELOW: pert_grid_f = get_minExtent_grid()
```
gmao-rreichle, line 635: @weiyuan-jiang: I assume this is not needed, but I'm not 100% sure.  It's possible that some fields of pert_grid_f that are set here are not filled in function get_minExtent_grid() below.  This may need further edits.  If so, then we should try to clean this up further, rather than just uncommenting this line again.

src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
```diff
@@ -688,11 +709,11 @@ subroutine Initialize(gc, import, export, clock, rc)
     tcinternal%pgrid_g = pert_grid_g
     tcinternal%pgrid_f = pert_grid_f
     tcinternal%pgrid_l = pert_grid_l
-    if(trim(grid_type) == "Cubed-Sphere" ) then
-      tcinternal%grid_g = tile_grid_g
-    else
-      tcinternal%grid_g = tcinternal%pgrid_g
-    endif
+    !if(trim(grid_type) == "Cubed-Sphere" ) then
+    tcinternal%tgrid_g = tile_grid_g
```
gmao-rreichle, line 713: @weiyuan-jiang: We shouldn't need the if block here, right?

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
```diff
@@ -1841,7 +1841,7 @@ subroutine RUN ( GC, IMPORT, EXPORT, CLOCK, RC )
          trim(out_path), trim(exp_id), exp_domain,                         &
          met_force, lai, cat_param, mwRTM_param,                           &
          tile_coord_l, tile_coord_rf,                                      &
-         tcinternal%pgrid_f, tcinternal%pgrid_g, tcinternal%grid_g,        &
+         tcinternal%tgrid_g, tcinternal%pgrid_f, tcinternal%pgrid_g,       &
```
gmao-rreichle, line 1844: @weiyuan-jiang: I changed the order of the arguments here -- it's more intuitive to have "tile_grid" before "pert_grid", at least to me.

src/Components/GEOSldas_GridComp/Shared/LDAS_TileCoordType.F90
```diff
@@ -161,10 +161,10 @@ module LDAS_TileCoordType
      type(tile_coord_type), pointer, contiguous :: tile_coord(:)=>null()
      type(tile_coord_type), pointer, contiguous :: tile_coord_f(:)=>null()
      integer, pointer    :: l2f(:)=>null()
-     type(grid_def_type) :: pgrid_g
-     type(grid_def_type) :: grid_g
-     type(grid_def_type) :: pgrid_f
-     type(grid_def_type) :: pgrid_l
+     type(grid_def_type) :: tgrid_g         ! tile_grid_g
```
gmao-rreichle, line 164: @weiyuan-jiang, I think the field name here should reflect that **tile**_grid aspect for clarity.

src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90
weiyuan-jiang: Without this if block, we broadcast the tile_grid_g twice when grid_type != "Cubed-Sphere".

gmao-rreichle: Yes, but I think it's the only clean way to assign tile_grid_g to tcinternal%tgrid_g here:

https://github.com/GEOS-ESM/GEOSldas/blob/f1922cf84ea2dc59ea55771cbae3862e2c8298ac/src/Components/GEOSldas_GridComp/GEOS_LdasGridComp.F90#L706

Before my changes, this was `tcinternal%[t]grid_g = tcinternal%pgrid_g`, which is somewhat backwards.  
`pgrid_g` is defined in get_pert_grid() based on `tgrid_g`.
While get_pert_grid() in its current version makes `pgrid_g` identical to `tgrid_g` unless we're in cube-sphere tile space, this could change in future revisions.  Then using `tcinternal%grid_g = tcinternal%pgrid_g` would also need to be changed, and that would be very easy to miss.  
I think we should stick to the logic that pgrid is derived from tgrid.  Broadcasting tile_grid_g should only require negligible resources.  It's just a broadcast (not a send/receive) and only a few strings and numbers.

src/Components/GEOSldas_GridComp/Shared/LDAS_TileCoordType.F90
```diff
@@ -214,7 +214,7 @@ function eq_grid_def_type( grid_1, grid_2 )
 
     type(grid_def_type), intent(in) :: grid_1, grid_2
     
-    if ( index(grid_1%gridtype,trim(grid_2%gridtype))>0        .and. &
+    if ( trim(grid_1%gridtype) == trim(grid_2%gridtype)        .and. &
```
gmao-rreichle, line 217: @weiyuan-jiang:  What's the motivation for this change?  I agree that this seems cleaner, but I'm wondering if the less stringent old formulation (i.e., `index(grid_1%gridtype,trim(grid_2%gridtype))>0` ) might have been chosen for a reason.  One example that should pass the old condition but not the new condition is the following:
```
grid_1%gridtype = SMAP_EASEv2
grid_2%gridtype = EASEv2
```
Could the change create problems with this kind of backward compatibility?

weiyuan-jiang: What if grid1 and grid2 exchange the position?  I think the original one is incomplete.

gmao-rreichle: @weiyuan-jiang: Yes, the old formulation isn't symmetric.  Perhaps it's just an accidental miss of "trim" in the first argument of the index function, but what if this was done intentionally for some reason?  I suppose we can always make the change and do the usual tests.  If they work, we should be ok. 
If we accept this change, what's still needed in order to merge this PR?  Have you or @biljanaorescanin tested it after my latest changes (i.e., after commit https://github.com/GEOS-ESM/GEOSldas/pull/637/commits/bc854c2d7e4664164a0325a9c41a49e8e55fdc1d)?

biljanaorescanin: I'm testing it now should have update for tomorrow.

biljanaorescanin:
All Nightly Tests passed on the branch.

gmao-rreichle (approved):

PR #640, opened 2023-03-23
fix problems and duplication with MATLAB tools for EASEv2 latlon2ind

gmao-rreichle:
PR fixes issues with **Matlab** tool for EASEv2 lat/lon-to-index mapping:

- Added optional input argument to `EASEv2_latlon2ind()` that controls rounding of [row, col] before returning. 
- Removed duplicate `smapeasev2_latlon2ind()` function from `write_smapL4SMqa.m`.

@amfox37, @gmao-qliu:  Please review the changes.   

cc: @biljanaorescanin


amfox37 (approved):
I've tested all but the mwRTM-related code, and things work as they should when using get_model_and_obs_clim_stats

PR #642, opened 2023-04-13
all processes have the same npert and pert_select

weiyuan-jiang:
@gmao-qliu Could you try this branch and see if this branch fixes your issue ? I will add some more checks on heterogenous std_dev and make sure it is greater than 0 on land.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
@@ -24,7 +24,7 @@ module LDAS_PertRoutinesMod
        nodata_generic,                            &
        nodata_tolfrac_generic,                    &
        nodata_tol_generic
-
+  use LDAS_ensdrv_mpi, only: mpicomm,numprocs,myid
   use MAPL_ConstantsMod,                ONLY:     &
        Tzero => MAPL_TICE,                        &
        alhe  => MAPL_ALHL,                        &
@@ -291,8 +291,6 @@ contains
     character(len=*), parameter :: Iam = 'read_ens_prop_inputs'
 
     ! MPI variables
-    type(ESMF_VM) :: vm
-    integer :: mpicomm
     logical :: root_proc,f_exist
 
     ! -----------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


gmao-qliu:
> @gmao-qliu Could you try this branch and see if this branch fixes your issue ? I will add some more checks on heterogenous std_dev and make sure it is greater than 0 on land.

Yes. The fix works in my tests.

src/Components/GEOSldas_GridComp/GEOSlandpert_GridComp/LDAS_PertRoutines.F90
```diff
@@ -1928,7 +1912,6 @@ subroutine get_pert_select( N_pert_max, pert_grid_l, std_pert, &
     pert_select(1:N_pert_max) = 0
 
     do k=1,N_pert_max
-
        do i=1,pert_grid_l%N_lon
           do j=1,pert_grid_l%N_lat
```
gmao-rreichle, line 1917: @weiyuan-jiang, thanks for adding the MPI_Allreduce, very elegant.   While I was looking at the code, I was wondering if we can/should eliminate the i and j loops here by using something like: 
`if (maxval( std_pert(k,:,:) ) > 0)  pert_select(k) = 1 `
Not sure if this executes faster than the double-loop, but it probably doesn't hurt.

weiyuan-jiang: Yes, that looks better

weiyuan-jiang:
@gmao-rreichle , I have added the check. I think it should be ready

weiyuan-jiang:
> > @gmao-rreichle , I have added the check. I think it should be ready
> 
> Thanks, @weiyuan-jiang. When I edited the intro to the PR so we have some description of what's done here, I started wondering if we actually need to enforce pert_std>0 for all grid cells that contain land tiles. Specifically, I now think that the MPI_AllReduce --> MPI_MAX operation may be just fine in dealing with a processor having exact zeros for pert_std in its subdomain when others have non-zero pert_std values. Say for some k we have one processor with pert_std(k,:,:)=0 but another processor with maxval(pert_std(k,:,:))>0, then after the MPI_AllReduce --> MPI_MAX operation we should end up with pert_select(k)=1 as desired. In other words, I don't think we actually need to force the user to avoid exact zero values where there is land. Put differently, the MPI_AllReduce --> MPI_MAX operation is a really nifty way of implementing what would otherwise be a fairly complicated communication+calculation. Does this make sense to you?

Yes, That makes sense to me. I can roll back the commit.

biljanaorescanin:
All our tests passed.

gmao-rreichle (approved):

PR #656, opened 2023-06-02
Adding ASCAT SM reader and matlab util for clim_stats

amfox37:
This branch contains: 

1. Code for reading and scaling 25km ASCAT soil moisture product
2. Matlab code for calculating climatology statistics to be used by the scaling on a regular 0.25 degree lat/lon grid and writing to a single netCDF file containing all 73 pentads.

The matlab code still needs minor attention as it contains some legacy of trying to be backwards compatible (which it isn't).

TO BE RESOLVED: 
The new BUFR reader for the ASCAT obs is **exceedingly slow** (see in subroutine `read_obs_sm_ASCAT_EUMET()` in `clsm_ensupd_read_obs.F90`).


src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,6 +1548,436 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+                 if ( datetime_lt_refdatetime( date_time_tmp, date_time_low ) .and.        &
+                      datetime_le_refdatetime( date_time_up, date_time_tmp )) cycle loop_report
+
+
+                  call ufbint(lnbufr,tmp_data,7,1,iret,'SSOM DOMO SMPF SMCF ALFR TPCX IWFR') 
+
+                
+                 ! skip if record contain no valid soil moisture value
+!                 call ufbint(lnbufr,tmp_data,1,1,iret,'SSOM')
+                 if(tmp_data(1) > 100. .or. tmp_data(1) < 0.) cycle loop_report
```
gmao-rreichle, line 1778: @amfox37, @gmao-qliu, @mathomp4: I'm not 100%, but it looks like these "while" loops read each obs+metadata separately.  If this is indeed the case, is there a way to read the entire file in one fell swoop and then apply the time window etc when we have everything in memory?

amfox37: Yes, I'd already tried consolidating the QC/flag reads (sorry for the nasty commented out lines in the draft) which didn't seem to impact things much, but could try consolidating the whole thing and then applying a QC loop.

amfox37:
@gmao-rreichle , @gmao-qliu, @mathomp4: As Rolf suggested, I went ahead and altered the code into a read of all subsets into memory, before then applying the QC. This gave a ~10% improvement in overall wallclock time. I then went down the rabbit hole of trying to remove the dependance on using the "SEC3" IO for openbf. This is required as these particular bufr files don't contain table information that would be self describing, so for each message section 3 needs to be interpreted with help from the bufr master tables and I thought that could be slow and isn't used anywhere else in the ADAS that I can see. 
To cut a long story short, I figured out how dump the internal dx bufr table generated by 'SEC3' IO, and modify it so it could be used with a more typical "IN" IO openbf call. But in the end this barely moved the dial on the timings, so I'm not including it here as it adds an extra somewhat esoteric file dependency.

gmao-rreichle:
Thanks, @amfox37, for changing the code to read the whole arrays into memory before applying QC.  It's disappointing that the improvement is only ~10%, which seems nowhere near enough.  This leaves the endian conversion as the next thing to try.  I don't recall the details of what @tclune and @mathomp4 had to say about this in the email chain, but I sincerely hope that by side-stepping the conversion of individual numbers we would get the improvement we need.  If there's a tool to convert the entire file from big-endian to little-endian, the first thing I'd try would be to convert the all BUFR files for the test case to little-endian and just re-run the test case.  If the files are little-endian upon input, we should be avoiding the loop that converts from big-endian to little-endian (as suggested by @gmao-qliu).  Maybe that'll do the trick.  Sorry this is such a pain.  
cc: @gmao-qliu @weiyuan-jiang

tclune:
Unfortunately, one cannot have a generic tool that can do the endian conversion in any easy manner.  This is because the conversion depends on whether a given item in the file is 4 byte or 8 byte.   If you know all data is one or the other, then it is easy to write a converter.   Failing that you need a customized application.  In theory you could have a separate file that itemizes 4 byte/8byte sequence for each record and a general tool would use that.  But i've never seen such a beast.

I'm somewhat doubtful that the endian conversion is the explanation.     We could try a separate application that simulates just the conversion of a comparable amount of data and measure the expense for comparison.

amfox37:
I have tried playing around with the open statement specifiers - specifically "convert=little_endian",(and =big_endian just to check...) , but I haven't found any impacts on timing.

src/Applications/LDAS_App/LDASsa_DEFAULT_inputs_ensupd.nml
```diff
@@ -2064,6 +2064,228 @@ obs_param_nml(48)%xcorr          = 0.1875
...
+obs_param_nml(49)%nodata         = -9999.
+obs_param_nml(49)%varname        = 'sfds'
+obs_param_nml(49)%units          = '%'
+obs_param_nml(49)%path           = '/discover/nobackup/projects/gmao/merra/iau/merra_land/DATA/ASCAT_EUMETSAT/Metop_A/'
+obs_param_nml(49)%name           = 'M02-ASCA-ASCSMO02'
+obs_param_nml(49)%scalepath      = ''
+obs_param_nml(49)%scalename      = ''
+obs_param_nml(49)%flistpath      = ''
+obs_param_nml(49)%flistname      = ''
+obs_param_nml(49)%errstd         = .04
```
gmao-rreichle, line 2096: @amfox37 : I think we should change this entry to be consistent with the units (%) listed above.  Fundamentally, the numbers here should all be consistent, so if the sfds data are in percent, then we should also have the errstd in %. I suppose it would be something like 9.0 (approx 0.04 m3/m3 / porosity * 100, with porosity = 0.45).  

The scaling routine for the ASCAT obs converts the obs value from % to m3/m3, and it should also convert the "errstd" number from % to m3/m3.  We may be able to derive a location-specific conversion factor from the info provided by the ASCAT scaling params.  Note that for AMSR-E sm retrievals, we also convert the input errstd (by multiplying with the ratio of the time series std-devs, as in:
`errstd_scaled = err_std * std_model / std_obs`
where std_model and std_obs are the time series std-devs of the model and obs, resp.  See subroutine scale_obs_sfmc_cdf(). This isn't perfect, but it is a rough way of accounting for the difference in the dynamic range of the model and obs (and, in the case of ASCAT, also for the difference in units).  I hope this makes sense.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,7 +1548,461 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+       tmpfname = fname_list(kk)
+
+       ! Are we in the required assimilation window?
+       !
+       ! e.g. Y2019/M07/M01-ASCA-ASCSMO02-NA-5.0-20190702075700.000000000Z-20190702084627-1350204.bfr
+       !
+       !      12345678901234567890123456789012345678901234567890123456789012345678901234567890 
+       !               1         2         3         4         5         6         7
+
+       str_date_time = tmpfname(40:53)
```
gmao-rreichle, line 1724: I don't know where the list files are and couldn't check the string processing here.  The sample "fname" probably needs to be updated.  Are the list files in day directories?  Or can/should we have the date info in the list file name?  This probably requires modifying the SMAP-specific subroutine that handles the list files.  (We may need this for the MODIS snow cover obs as well.)  Would it help to have the BUFR obs files in day directories?

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,7 +1548,461 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+    
+    ! --------------------
+    !
+    ! variables for BUFR read
+
+    real*8, dimension(15) :: tmp_vdata
+
+    integer, parameter :: lnbufr  =     50       ! BUFR file unit number
+    integer, parameter :: max_rec =  20000       ! max number of obs after QC
+    integer, parameter :: max_obs = 250000       ! max number of obs read by subroutine
```
gmao-rreichle, line 1634: I don't think there is a check if we exceed `max_obs`. If that ever happens, we probably get a nasty memory error.  Probably needs to be added.
A good value for `max_rec` depends on `dtstep_assim` and on whether we're reading just one orbit direction.  Here, too, we'll probably need a check.  Possibly we can make `max_rec` depend on `dtstep_assim` if we want to minimize memory demand, although that's probably not a priority now.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,7 +1548,461 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+      
+       read(str_date_time( 1: 4), *) date_time_tmp%year
+       read(str_date_time( 5: 6), *) date_time_tmp%month
+       read(str_date_time( 7: 8), *) date_time_tmp%day
+       read(str_date_time( 9:10), *) date_time_tmp%hour
+       read(str_date_time(11:12), *) date_time_tmp%min
+       read(str_date_time(13:14), *) date_time_tmp%sec
+       
+       if ( datetime_lt_refdatetime( date_time_low_fname, date_time_tmp ) .and.          &
+            datetime_le_refdatetime( date_time_tmp,       date_time_up  )       ) then
```
gmao-rreichle, line 1734: Are the "lt" and "le" comparisons consistent with the half-open interval `[date_time_low, date_time_up)` mentioned above?  Should this be "le"  first and then "lt"?

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,7 +1548,461 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+       
+       do kk = 1,N_files
+          
+          ! open bufr file
+          
+          call closbf(lnbufr)  ! if a file with unit number lnbufr is open in (or "linked" with) BUFR, close it
+          open(lnbufr, file=trim(fnames(kk)), action='read', form='unformatted')
+          call openbf(lnbufr, 'SEC3', lnbufr) 
+          call mtinfo( trim(this_obs_param%path) // '/BUFR_mastertable/', lnbufr+1, lnbufr+2)
+          call datelen(10)
```
gmao-rreichle, line 1780: A couple of comments/questions here:
1. I looked up the "openbf()" interface for [NCEPLIBS-bufr 12.0.1](https://noaa-emc.github.io/NCEPLIBS-bufr/openbf_8f.html).  The doc there says that the unitnumbers passed into openbf() must already be associated with fortran-opened files, which I don't think they are.  I suspect the library we're using now Fortran-opens the file at the same time as it is being linked with the BUFR lib, and the Fortran-open call is no longer needed, but I'm only guessing.  However, if that's the case, why do we Fortran-close the file below, as in `close(lnbufr)` ?
2. Why do we use 'SEC3' mode, which "use[s] Section 3 of input messages for decoding rather than DX BUFR table information from LUNDX" (per the doc above).  Don't we have the BUFR tables?  Since we loop through each individual record here, asking the library to figure out the BUFR table on each read may add a lot of overhead.  But here, too, I'm only guessing.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -1548,7 +1548,461 @@ subroutine read_obs_sm_ASCAT(                                  &
...
+          if ( tmp_data(kk, 7) > 100. .or. tmp_data(kk, 7) < 0. ) cycle     
+          
+          ! EUMETSAT file contains data of both ascending and descending half-orbits. 
+          ! DOMO ("Direction of motion of moving observing platform") is used to separate Asc and Desc
+          !   because the file doesn't contain any explicit orbit indicator variable.
+          ! According to Pamela Schoebel-Pattiselanno, EUMETSAT User Services Helpdesk:
+          !   "When the value (of DOMO) is between 180 and 270 degrees, it is the descending part 
+          !    of the orbit.  When it is between 270 and 360 degrees, it is the ascending part."
+          if (index(this_obs_param%descr,'_A') /=0 .and. (tmp_data(kk, 8) < 270 .or. tmp_data(kk, 8) >  360)) cycle
+          if (index(this_obs_param%descr,'_D') /=0 .and. (tmp_data(kk, 8) < 180 .or. tmp_data(kk, 8) >= 270)) cycle
```
gmao-rreichle, line 1831: Maybe let's revisit the separation of the ASCAT sm species into Asc and Desc.  In the past, we haven't really used this distinction in the LDAS for anything, and it just doubles the total cost of the BUFR read.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
gmao-rreichle: Please disregard item 1.  Apologies, I somehow missed the Fortran open() just before the openbf().  After discussion, we can also disregard item 2.

src/Applications/LDAS_App/LDASsa_DEFAULT_inputs_ensupd.nml
gmao-rreichle: Apologies, I missed that the errstd is already rescaled in `scale_obs_sfmc_zscore()`.  The errstd entry in the nml file still needs changing, though.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
amfox37: Fixed comment to reflect what we're actually doing with the half-open interval

https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L1668C1-L1669C60

amfox37: Change half-open interval comment
```
    ! find files that are within half-open interval 
    !   (date_time-dtstep_assim/2,date_time+dtstep_assim/2]
```

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
amfox37: This was a hack, fixed by:
- read_obs_SMAP_fnames -> [read_obs_fnames](https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L7149)
- optional argument [obs_dir_hier](https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L7174C1-L7174C66) in read_obs_fnames to write YYYY/MM/filename
```fortran
          if (present(obs_dir_hier) .and. obs_dir_hier == 1) then
            
            ! preface file names with "Yyyyy/Mmm"
            
            fname_list(ii) = YYYYMMdir // trim(tmpstr80)


          elseif (present(obs_dir_hier) .and. obs_dir_hier /= 1) then
            
            err_msg = 'Unrecognized obs_dir_hier #'
            call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)


          else
            ! preface file names with "Yyyyy/Mmm/Ddd" Default for SMAP obs
            
            fname_list(ii) = YYYYMMDDdir // trim(tmpstr80)
          
          end if
```
- update date from string [parse]( https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L1726C8-L1726C39) to reflect this change

src/Applications/LDAS_App/LDASsa_DEFAULT_inputs_ensupd.nml
amfox37: add /100. after initial read of [errstd](https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L1954) during tile super-obbing
```fortran
             ASCAT_sm_std(ii) = this_obs_param%errstd/100.    ! change units from percent (0-100) to fraction (0-1)
```

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
amfox37: add [error out](https://github.com/GEOS-ESM/GEOSldas/blob/3f95ee02780cf30259b9a2ea85eceb000822a7e7/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90#L1786C1-L1789C23) if maxobs exceeded
```fortran
                if (N_obs > max_obs) then
                  err_msg = 'Attempting to read too many obs - how long is your assimilation window?'
                  call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)
                end if
```

amfox37: Still need to do the same with N_tmp and max_rec

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -7657,6 +8165,199 @@ subroutine scale_obs_tskin_zscore( N_catd, tile_coord,   &
...
+    real,    intent(in),    dimension(N_catd) :: tmp_lon, tmp_lat
+    real*8,  intent(in),    dimension(N_catd) :: tmp_time         ! J2000 seconds
+        
+    ! inout
+    
+    real,    intent(inout), dimension(N_catd) :: tmp_obs
+    real,    intent(inout), dimension(N_catd) :: tmp_std_obs
+    
+    ! ----------------------------------------------------------
+    ! Grid and netcdf parameters (might want to read these from netCDF file?)
```
gmao-rreichle, line 8204: Yes, I think that's better. Or, at the very least, we need to check that the N_lat and N_lon that are read from the file are consistent with the hardcoded dlat and dlon (and - if that can be done sensibly - also ll_lat and ll_lon).

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -7657,6 +8165,199 @@ subroutine scale_obs_tskin_zscore( N_catd, tile_coord,   &
...
+    real,    parameter :: ll_lon  = -180.0000
+    real,    parameter :: ll_lat  =  -90.0000 
+    real,    parameter :: dlon    =    0.25
+    real,    parameter :: dlat    =    0.25
+    
+    ! ----------------------------------------------------------
+    
+    ! local variables
+        
+    real, parameter :: tol = 0.99
```
gmao-rreichle, line 8217: "tol" is in units of deg lat-lon, right? That would mean we extrapolate the scaling parameters up to ~100 km out at lower latitudes.  For longitude checks at high lats we'd want larger values of tol.  This might need more thought.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -7657,6 +8165,199 @@ subroutine scale_obs_tskin_zscore( N_catd, tile_coord,   &
     
   end subroutine scale_obs_tskin_zscore
 
+ 
+  ! *****************************************************************
+  
+  subroutine scale_obs_sfmc_zscore( N_catd, tile_coord,         &
```
gmao-rreichle, line 8171: A reminder that we need to deal with unphysical scaled soil moisture values -- could add as a new check in get_obs_pred(), but must be separate from qc_model(), which only disables obs by setting obs_pred to no-data.  would need to add functionality that computes obs-space porosity and then cuts back scaled soil moisture obs to obs-space porosity at the high end and to zero at the low end

mathomp4:
Per my email, I think if NCEP_Shared is moved to v1.3.0 that should fix the GNU build issue.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_latlon_grid.m
```diff
@@ -0,0 +1,351 @@
...
+        % read innov files
+        fname = [inpath, '/ana/ens_avg/', 'Y', YYYYMMDD(1:4), '/', 'M', YYYYMMDD(5:6), '/', exp_run, '.ens_avg.ldas_ObsFcstAna.', YYYYMMDD, '_', HHMM, 'z.bin'];
+        ifp = fopen(fname, 'r', 'l');
+        
+        if (ifp > 0) % Proceed only if file exists
+          fclose(ifp);
+          [date_time, obs_assim, obs_species, obs_tilenum, obs_lon, obs_lat, obs_obs, obs_obsvar, obs_fcst, obs_fcstvar, obs_ana, obs_anavar] = read_ObsFcstAna(fname);
+          
+          % remove tiles when there is no obs_fcst (obs_fcst == 0 in innov output when missing)
+          idx = find(obs_fcst == 0);
```
gmao-rreichle, line 161: @gmao-qliu @amfox37: 
Not sure why the check here is for "obs_fcst==0".  Do we not have a defined no-data-value for obs_fcst?  The zero value might work for ASCAT sfds, but it's easy to imagine that zero is problematic if in future we're trying to generalize the subroutine for other variables (e.g., SWE).  
My guess is that obs_fcst is -9999 (SMAP no-data-value) or possibly 1.e15 (MAPL_UNDEF).
Having said that, we probably don't even need to keep this block.  The matlab reader for "ObsFcstAna" files produces NaN for no-data.  I think this is inherited from "get_model_and_obs_clim_stats.m" and would need to be changed there as well.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/write_netcdf_latlon_grid.m
```diff
@@ -0,0 +1,246 @@
...
+    netcdf.putAtt(ncid, varid_lon,         'standard_name', 'longitude');
+    netcdf.putAtt(ncid, varid_lon,         'long_name',     'lower left longitude of gridcell');
+    netcdf.putAtt(ncid, varid_lon,         'units',         'degrees_east');
+    netcdf.putAtt(ncid, varid_lon,         'axis',          'X');
+    
+    varid_lat        = netcdf.defVar(ncid, 'lat',        float_precision, [dimid_lat]);
+    netcdf.putAtt(ncid, varid_lat,         'standard_name', 'latitude');
+    netcdf.putAtt(ncid, varid_lat,         'long_name',     'lower left latitude of gridcell');
+    netcdf.putAtt(ncid, varid_lat,         'units',         'degrees_north');
+    netcdf.putAtt(ncid, varid_lat,         'axis',          'Y');
```
gmao-rreichle, line 139: @amfox37 : The variable names 'lon' and 'lat' here are not good practice.  They are really 'll_lons' and 'll_lats', but the 'll_*' bit is not reflected in the variable name.  If someone just looks at the nc file and doesn't check the long_name, they might not get this.  
The grid here is a regular lat/lon grid, and we already have the scalar variables 'll_lon' and 'll_lat' in the file.  Together with 'd_lon', 'd_lat', 'N_lon', and 'N_lat' we can readily construct 'll_lons' and 'll_lats'.  
The simplest and safest thing to do is probably to not write 'lon' and 'lat' into the scaling parameter file.  
Alternatively, we could rename the variables to 'll_lons' and 'll_lats'.
Or we could write the grid cell center 'lon' and 'lat' values (i.e., write lon=ll_lons+d_lon/2 and lat=ll_lats+d_lat/2).

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -7645,6 +8156,283 @@ subroutine scale_obs_tskin_zscore( N_catd, tile_coord,   &
...
+    !  int pentad(pentad) ;
+    !          pentad:standard_name = "pentad" ;
+    !  double start_time(pentad) ;
+    !          start_time:standard_name = "start time" ;
+    !  double end_time(pentad) ;
+    !          end_time:standard_name = "end time" ;
+    !  double lon(lon) ;
+    !          lon:standard_name = "longitude" ;
+    !  double lat(lat) ;
+    !          lat:standard_name = "latitude" ;
```
gmao-rreichle, line 8190: This is where the meaning of 'lon' and 'lat' as the **lower** **boundaries** of the grid cells is lost.  See my comment re. `write_netcdf_latlon_grid.m`

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
```diff
@@ -7645,6 +8156,283 @@ subroutine scale_obs_tskin_zscore( N_catd, tile_coord,   &
...
+          
+          ! Find indices for current tile lat and lon on scaling parameter grid
+
+          i_ind = ceiling((this_lon - ll_lon)/dlon) 
+          j_ind = ceiling((this_lat - ll_lat)/dlat) 
+          
+          ! Sanity check (against accidental use of wrong tile space)
+
+          if ( abs(tile_coord(i)%com_lat-sclprm_lat(j_ind))>tol  .or.             &
+               abs(tile_coord(i)%com_lon-sclprm_lon(i_ind))>tol ) then
```
gmao-rreichle, line 8377: This check compares the tile's center-of-mass with the  **lower boundary** of the scaling parameter grid cell.  It would make more sense to have 'lon' and 'lat' indicate the lon & lat of the **center** of the scaling parameter grid cells.  I think this is the only place that uses 'lon' and 'lat' from the netcdf scaling param file.  We may not even need this check.

gmao-rreichle:
@amfox37 @gmao-qliu : 
I cleaned up the code (mostly matlab) a bit more.  Here's the summary diff of my changes (excl white space changes):
https://github.com/GEOS-ESM/GEOSldas/compare/c3fa251..4113033?w=1
Included in the above is the replacement of the deprecated nanmean, nanstd, and nansum functions in this single commit:
https://github.com/GEOS-ESM/GEOSldas/pull/656/commits/3ffee32c6b6bc092d9ea74200de33494ba0ed891
I also merged two lines into one line here: 
https://github.com/GEOS-ESM/GEOSldas/blob/411303308f67a7de5b7113cbba7b098a695a35ea/src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_latlon_grid.m#L238
Please inspect the changes and test that the revised matlab processing scripts still work as expected.

amfox37:
> Please inspect the changes and test that the revised matlab processing scripts still work as expected.

OK, will check it reproduces existing scaling parameters. 

Not sure what to do for the best with the lat/lon thing. Seems like I should change to write center lat/lon to the netCDF file, keep that name and change long_name, or just forget the check which is inherited from elsewhere, and maybe not required and so not write these lat/lon at all.

gmao-rreichle:
> Not sure what to do for the best with the lat/lon thing. Seems like I should change to write center lat/lon to the netCDF file, keep that name and change long_name, or just forget the check which is inherited from elsewhere, and maybe not required and so not write these lat/lon at all.

Maybe best to drop 'lat' and 'lon' from the scaling param netcdf file because these vectors are redundant, after all.  We could still keep the sanity check in the zscore scaling subroutine by computing the (grid-cell center) sclprm_lat/lon coordinates on the fly (from ll_lat, ll_lon, d_lat, d_lon, N_lat, and N_lon) instead of reading them from the netcdf file.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
amfox37: So having looked at where tmp_lon and tmp_lat come from it doesn't seem like we really need to keep this check. 

There is definitely some circularity as although tmp_lon and tmp_lat are derived solely from observation values, the observations themselves are initially assigned to a tile based on `tile_coord%com_lat` or `tile_coord%com_lon` in `get_tile_num_from_latlon`, so I don't really see how this test could ever fail. I guess its definitely checking nothing has gone wrong in the "super ob-ing" stage.

But if we want to keep it we wouldn't need to use the lat and lon vectors read from the netCDF file, we could just calculate it at the time given the grid info.
Something like: 
         
```
          this_lon = tmp_lon(i)
          this_lat = tmp_lat(i)
          
          ! Find indices for current tile lat and lon on scaling parameter grid

          i_ind = ceiling((this_lon - ll_lon)/dlon) 
          j_ind = ceiling((this_lat - ll_lat)/dlat) 

          sclprm_lon = (ll_lon - dlon/2) + (i_ind*dlon)
          sclprm_lat = (ll_lat - dlat/2) + (j_ind*dlat)
                    
          ! Sanity check (against accidental use of wrong tile space)

          if ( abs(tile_coord(i)%com_lat-sclprm_lat)>tol  .or.             &
               abs(tile_coord(i)%com_lon-sclprm_lon)>tol ) then
                 
               err_msg = 'Lat/lon diff beyond tolerance'
               call ldas_abort(LDAS_GENERIC_ERROR, Iam, err_msg)

          end if
```

gmao-rreichle: Thanks, @amfox37, for taking another look at this.  I agree with the assessment.  When you get a chance, please remove the sanity check and remove lat/lon from the scaling parameter file.

src/Applications/LDAS_App/util/inputs/obs_scaling_params/get_model_and_obs_clim_stats_latlon_grid.m
gmao-rreichle: @amfox37, @gmao-qliu : With any luck, my commit https://github.com/GEOS-ESM/GEOSldas/pull/656/commits/565c8682e3c1aead0dfc7263e89eee4110336138 fixes this in both `get_model_and_obs_clim_stats*.m` scripts.  Please let me know if you see any problems.

src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/clsm_ensupd_read_obs.F90
amfox37: @gmao-rreichle  Ok, have done this in commit [544a119](https://github.com/GEOS-ESM/GEOSldas/pull/656/commits/544a119cdf58f83b37204028c97eeea77d9ce9c5) removing lat/lon matlab write, lat/lon FORTRAN read and sanity check in read_obs.
Basic testing suggest this change, and all other changes in matlab scripts are good to go.

gmao-rreichle (approved):
Passed full suite of GEOSldas nightly tests (@gmao-rreichle, 24 Dec 2023)

PR #657, opened 2023-06-22
Hotfix: Change to GEOS_Util v2.0.1 (components.yaml)

gmao-rreichle:
Updating to GEOS_Util v2.0.1 to fix a problem with remapping restarts in GEOSldas v17.13.0.
This is a hot fix directly on top of GEOSldas v17.13.0 to avoid the non-zero diff change associated with ESMA_env v4.16.0, which is already on the GEOSldas "develop" branch.
The bug fix in GEOS_Util is documented here: https://github.com/GEOS-ESM/GEOS_Util/pull/23 
@biljanaorescanin: Testing this PR is tricky because the current BASELINE in the nightly tests uses ESMA_env v4.16.0, which is not 0-diff against v17.13.0.  It should be possible to just run the nightly tests for this PR using the usual script, but then verify (manually) against the PREVIOUS test output (i.e., the older baseline).  If the latter is no longer available for some reason, we'd have to create output using v17.13.0.  The key is to make sure this PR is zero-diff against v17.13.0.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -25,7 +25,7 @@ GMAO_Shared:
 GEOS_Util:
   local: ./src/Shared/@GMAO_Shared/@GEOS_Util
   remote: ../GEOS_Util.git
-  tag: v2.0.0
+  tag: v2.0.1
 
 MAPL:
   local: ./src/Shared/@MAPL


biljanaorescanin:
All Nightly Tests passed except cube sphere test cases (globalcs aggglobalcs).  
In specific tests failed for output comparison in model runs for "lnd" and "lfs" collections and in assim run for "inst3_2d_lndfcstana" collection.  
Our oldest option for comparison is "baseline" from May30th.  Differences are all roundoff but unfortunately "baseline" that is oldest is from after the non zero diff change (https://github.com/GEOS-ESM/GEOSldas/pull/654).  I'm running v17.13.0. (with ESMA v4.9.1) that is what we had before the non zero diff merge. Once I compare "cs" cases I will add new comment.

biljanaorescanin:
When we run nightly tests on the "main" we get same fails for globalcs and  aggglobalcs cube sphere test cases as on this branch. I can confirm comparisons are zero diff for both model and assim fails we see.

gmao-rreichle:
Thanks for the tests, @biljanaorescanin!  When you get a chance, please approve

biljanaorescanin (approved):

PR #663, opened 2023-06-28
CO2 for CatchCN

biljanaorescanin:
This CO2 file is needed for CatchCN offline simulations, it is not resolution dependent. 
It is now stored within bcs versions. It is in `legacy_bcs` for versions @gmao-jkolassa is running.


biljanaorescanin:
I've tested ldas_setup; and it is properly linked once we run for experiment setup.

gmao-rreichle (approved):

PR #669, opened 2023-07-06
Move GEOSldas.F90 to use fArgParse

mathomp4:
This PR moves GEOSldas.F90 to use fArgParse instead of FLAP. In MAPL 3, FLAP support will be removed, but we can transition to fArgParse now that we use Baselibs 7.13.0. 

This is similar to PRs made for other executables (see https://github.com/GEOS-ESM/GEOSgcm_App/pull/441)

This should work, though I'm sure @biljanaorescanin will test. But this isn't a critical fix, but we might as well transition from FLAP sooner rather than later.

Note: fArgParse doesn't (yet?) support the description and author fields when making the CLI.

### DIFF
--- a/src/Applications/LDAS_App/GEOSldas.F90
+++ b/src/Applications/LDAS_App/GEOSldas.F90
@@ -4,7 +4,7 @@
 
 program LDAS_Main
 
-  
+
    ! !USES:
    use MAPL
    use GEOS_LDASGridCompMod, only:  ROOT_SetServices => SetServices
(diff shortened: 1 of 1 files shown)


gmao-rreichle:
Thanks, @mathomp4!  Looks good to me.  @biljanaorescanin, please test when you get a chance, just to be safe.  It doesn't look like we'll need the full set of LDAS tests.  If this works for one of the Intel [or GNU] tests, it should work for all of them.  So maybe just run LDAS_CONUS and LDAS_GNUCONUS.  Thanks.

biljanaorescanin:
We passed tests.

gmao-rreichle (approved):

PR #681, opened 2023-10-26
Update Repositories

biljanaorescanin:
This PR is to update repos in components.yaml file to newer versions and to add option to ldas_setup to use either cas or sky.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,18 +47,6 @@ To obtain a build that is suitable for debugging, use `parallel_build.csh -debug
 
 See below for how to build the model in multiple steps.
 
-#### Running on Milans at NCCS
-
-By default, `parallel_build.csh` will build on SLES12 nodes at discover (either Skylake or Cascade Lake). However, Milan nodes were
-recently installed at discover, which use SLES15 as their operating system. Because of this OS difference, if you want to run on the
-Milans, you must build on the Milans. To do so you can run `parallel_build.csh` with the `-mil` flag, e.g.:
-
-```
-parallel_build.csh -mil
-```
-
-This will by default build in `build-SLES15` and install to `install-SLES15`.
-
 ---
 
 ## How to Set Up (Configure) and Run GEOSldas
--- a/README.md
+++ b/README.md
@@ -4,7 +4,7 @@ This document explains how to build, set up, and run the GEOS land modeling and 
 
 ## How to Build GEOSldas
 
-### Step 1: Load the Build Modules  
+### Step 1: Load the Build Modules
 
 Load the `GEOSenv` module provided by the GMAO Software Infrastructure team.  It contains the latest `git`, `CMake`, and `mepo` modules and must be loaded in any interactive window that is used to check out and build the model.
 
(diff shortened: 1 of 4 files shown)


gmao-rreichle:
@biljanaorescanin, @mathomp4 : After seeing Matt's edits of README.md, I now realize that [my latest commit isn't going to work](https://github.com/GEOS-ESM/GEOSldas/pull/681/commits/c911d23b24e65812def818001bcfbaf554678e88).  
I guess the idea of the preceding commit by Matt would have the user to choose Milan via the n-tasks-per-node setting.  That should work, but we might need more documentation in the batinp file to explain the SLES12 (cas, sky) vs. SLES15 (mil) difference.  Thoughts?

mathomp4:
> @biljanaorescanin, @mathomp4 : After seeing Matt's edits of README.md, I now realize that [my latest commit isn't going to work](https://github.com/GEOS-ESM/GEOSldas/pull/681/commits/c911d23b24e65812def818001bcfbaf554678e88). I guess the idea of the preceding commit by Matt would have the user to choose Milan via the n-tasks-per-node setting. That should work, but we might need more documentation in the batinp file to explain the SLES12 (cas, sky) vs. SLES15 (mil) difference. Thoughts?

@gmao-rreichle Yes. You can't do that. In the GCM I use the fact that CMake knows where you build to "lock" a user into what they can run on:

https://github.com/GEOS-ESM/GEOSgcm_App/blob/develop/gcm_setup#L439

At CMake time, cmake fills in the `BUILT_ON_SLES15` variable with `TRUE` or `FALSE`. My guess is working with @weiyuan-jiang it should be possible to do something similar with `ldas_setup` by maybe throwing an Error if a user tries to use a SLES12 build on SLES15 and vice versa.

mathomp4:
Also it might be a while before I figure out how to do nightly testing scripts to work on Milans. The main issue is that `discover-cron` is SLES12 and some bits of the scripting run there while other bits run on the compute nodes you get. That mix of OSs causes nightmares (see [5 releases of ESMA_env in 3 days](https://github.com/GEOS-ESM/ESMA_env/releases))

Disentangling that will be a challenge. I don't think impossible, but a challenge. For now, I'd recommend using those scripts only on SLES12.

biljanaorescanin:
@mathomp4 its successfully compiled with `./parallel_build.csh -mil` it created two directories `build-SLES15` and `install-SLES15`. 
@gmao-rreichle I will remove added constrain for `milan` you added.

biljanaorescanin:
We pass all Nightly Tests.

tclune (approved):
infrastructure changes ok

gmao-rreichle (approved):
I verified that the comparison failure of the AGGGLOBALCS/assim test is just roundoff, as expected.  Approving.

README.md
```diff
@@ -57,7 +69,7 @@ a) Set up the job as follows:
 cd (build_path)/GEOSldas/install/bin
```
weiyuan-jiang, line 57: Is this correct?

gmao-rreichle: I suppose this would need to be something like cd (build_path)/GEOSldas/install-SLES12/bin
Let's fix it when we include the SLES15 vs SLES12 instructions in #693

PR #685, opened 2023-11-23
Remove requirement for mwRTM parameter files inputs

amfox37:
_Current behavior_
Currently the LDAS is assuming mwRTM parameter files inputs will be available and used, even if the mwRTM isn't being used in the analysis or for producing outputs (such as in an ASCAT SM only analysis). It exits trying to read a nonexistent vegopacity file, or passing an un-initialized `mwRTM_param` variable to `get_enkf_increments`. 

_Potential fix_
In GEOS_LandAssimGridComp.F90
1. `get_vegopcity` and `mwRTM_param_nodata_check` are only called if mwRTM is set to be true, which it will be if there mwRTM files in the nml.
2. If `mwRTM_param` isn't allocated during a mwRTM parameter file read, then `allocate(mwRTM_param(1))` just before call to `get_enkf_increments`

(It's unclear to me if allocating `mwRTM_param` like this but not  giving a value to it is good/acceptable practise)

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/GEOS_LandAssimGridComp.F90
@@ -1886,19 +1886,29 @@ contains
          )
     _VERIFY(status)   
 
-    ! mwRTM_param already contains static parameters, only need vegopacity.
+    if ( mwRTM ) then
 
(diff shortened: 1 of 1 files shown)


amfox37:
@gmao-rreichle @weiyuan-jiang  I have checked ASCAT only assimilation with the 4 Intel/GNU and regular/aggressive compiler/optimization  combinations and `allocate(mwRTM_param(0))`  and all seems good with this change which I just commited

PR #688, opened 2023-12-04
add the capability to read the GEOSIT forcing

gmao-qliu:
I edited the file so GEOSldas can use the GEOSIT lat/lon output files (w/wo corrected precipitation) as forcing.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -3170,6 +3170,7 @@ contains
     real,    parameter :: nodata_GEOSgcm = 1.e15   
     
     character(40), dimension(N_G5DAS_vars,              N_defs_cols) :: G5DAS_defs
+    character(40), dimension(N_G5DAS_vars,              N_defs_cols) :: GEOSIT_defs
     character(40), dimension(N_MERRA_vars,              N_defs_cols) :: MERRA_defs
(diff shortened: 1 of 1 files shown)


src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -4957,6 +5018,23 @@ subroutine parse_G5DAS_met_tag( met_path_in, met_tag_in, date_time, &
 
        end if
 
+    elseif (met_tag_out(1:18)=='cross_d5924_GEOSIT') then
```
gmao-rreichle, line 5021: @gmao-qliu : for consistency with RPIT/FPIT, I think we need to keep the GEOS version in the "cross" stream definition. As was the case with RPIT/FPIT, a future version of GEOS-IT might be produced with a newer GEOS system but still under the "GEOS-IT" name.  This change requires you do edit the MET_TAG in your LDAS.rc file.

gmao-rreichle:
@gmao-qliu : I added one commit with a few minor changes:
https://github.com/GEOS-ESM/GEOSldas/pull/688/commits/51b900d6d121710fb9f2b644fa5fd032b551a149?diff=split&w=1
Except for the tweak of the GEOS-IT "cross" stream MET TAG (see comment above), the edits should not change anything and be zero-diff from the previous commit.   Please take a look and try a brief simulation to verify.  Thanks!

gmao-rreichle (approved):
successfully 0-diff tested for full suite of GEOSldas nightly tests on 6 and 7 Dec 2023

PR #718, opened 2024-02-21
support reading corrected precipitation from aggregated daily netcdf files

gmao-qliu:
To reduce the number of files on disk, we can aggregate the 24 hourly corrected precipitation files into 1 daily file.   The updates support reading from both hourly and daily (if no hourly file is found) files.

### DIFF
--- a/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
+++ b/src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
@@ -3207,7 +3207,7 @@ contains
 
     logical :: minimize_shift, use_prec_corr, use_Predictor, tmp_init
 
-    logical :: daily_met_files
+    logical :: daily_met_files, daily_precipcorr_files
     
     integer :: nv_id, ierr, icount(3), istart(3), lonid, latid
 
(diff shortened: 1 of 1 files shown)


src/Components/GEOSldas_GridComp/GEOSmetforce_GridComp/LDAS_Forcing.F90
```diff
@@ -5691,8 +5692,20 @@ subroutine get_GEOS_corr_prec_filename(fname_full,file_exists, date_time, met_pa
     
     fname_full_tmp1 = trim(fname_full)                 ! remember for error log below
     
+    ! second try: look for daily file in year/month dir 
+    is = index(fname,'z.nc')
```
gmao-rreichle, line 5696: @gmao-qliu : I think this only works if met_tag does not include "GEOS.fp.asm".  When met_tag includes "GEOS.fp.asm", then fname ends in "_[HHMM].V01.[file_ext].   See if block starting in line 5658 above.  In this case, determining "is" from the position of "z.nc" in the "fname" string won't work.  Since we know the date & time, a better strategy would be start with "fname" and replace [YYYYMMDD_HHMM] with [YYYYMMDD] to construct "fname_tmp"?   Or simply construct fname_tmp in addition to fname within the if block above (line ~5658)?   Having said that, do we even have corrected precip files that use the "GEOS.fp.asm.[...].V01.[file_ext]" template?  In any case, this probably needs more thought and probably a correction.

gmao-qliu: that's good point. Yes we use the GEOS.fp.asm*V01.nc4 format in OPS. Replacing fname would work.

gmao-qliu: "simply construct fname_tmp in addition to fname within the if block above (line ~5658)"   I like this option better as the "replace" one also needs to deal with "z.nc"  and "V01.nc" separately.

gmao-rreichle:
I successfully tested the branch after my minor updates.  @gmao-qliu, did you verify that the new functionality of reading from daily files still works after your most recent changes?

gmao-qliu:
> I successfully tested the branch after my minor updates. @gmao-qliu, did you verify that the new functionality of reading from daily files still works after your most recent changes?

Yes. Verified.

gmao-rreichle (approved):

Issue #713, opened 2024-02-16
Testing Baselibs 7.18.1 → HDF5 1.14


mathomp4:
Soon ESMA_env will move to use Baselibs 7.18.1 whose main change is moving from HDF5 1.10 to 1.14. GEOSgcm seems pretty agnostic to this as any HDF5 work is done through netCDF (solely I think).

But GEOSldas has [an entire file based on HDF5](https://github.com/GEOS-ESM/GEOSldas/blob/develop/src/Components/GEOSldas_GridComp/GEOSlandassim_GridComp/io_hdf5.F90), so we want to be careful.

As such, I've made a branch here:

```
feature/mathomp4/baselibs-7.18.1-ldas
```

which updates the ESMA_env to use Baselibs 7.18.1. 

I'm hoping nothing happens, all works well, and life is good. 🤞🏼 

See also #701

gmao-rreichle:
The "CONUS" and "GLOBAL" tests passed for both Intel and GNU.  The GLOBAL/assim case is akin to running GEOSldas.x within the SMAP L4_SM system and exercises the reading of h5 files. 
The SMAP L4 ops system has additional uses of hdf (e.g., creating the h5 product files for L4_SM and L4_C), although I'm not sure if @jardizzo uses ESMA_env to build the other components of the SMAP L4_SM system (other than GEOSldas).  I'll inquire.

mathomp4:
@gmao-rreichle Well, glad to hear GEOSldas can at least *consume* HDF5 just fine. My guess is if SMAP L4 does use Baselibs, they probably have no reason or need to use newer ones. So we could be good!

gmao-rreichle:
Closing this issue.  Discussion continues in #726.

PR #725, opened 2024-02-29
Update utilities versions (components.yaml)

gmao-rreichle:
This PR updates GEOSldas to the latest versions of the utilities in preparation for next release:
```
ESMA_cmake  v3.41.0
GMAO_Shared v1.9.7
GEOS_Util   v2.0.7
MAPL        v2.44.0
```

Note: ESMA_env v4.26.0 does **not** work, see #726 

0-diff tests by @biljanaorescanin all passed.  

Please let me know if you have any concerns re. the above versions. 

@biljanaorescanin @weiyuan-jiang @mathomp4

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -5,7 +5,7 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v4.26.0
+  tag: v4.23.0
   develop: main
 
 cmake:
--- a/components.yaml
+++ b/components.yaml
@@ -5,13 +5,13 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v4.23.0
+  tag: v4.26.0
   develop: main
 
 cmake:
   local: ./@cmake
   remote: ../ESMA_cmake.git
-  tag: v3.40.0
+  tag: v3.41.0
   develop: develop
 
 ecbuild:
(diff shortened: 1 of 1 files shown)


mathomp4:
I think those are all good to me. 

Note that eventually I can work with @weiyuan-jiang to add the (new style) MPI detection into `lenkf.j` so that you can set MPI flags/variables depending on the stack (like how the GCM does for Intel MPI, Open MPI, MPT...)

gmao-rreichle:
> I think those are all good to me

Thanks, @mathomp4.  Alas, I'm getting a floating point exception early in the LDAS_DEBUGCONUS/model run.  The GEOSldas log file (./scratch/GEOSldas_log_txt) suggests it may be in the forcing initialization, but unfortunately it's not clear immediately where the crash occurs.  Interestingly, GNUDEBUGCONUS runs fine. 
I'm not sure what in the new utilities may be causing this.  I can try to roll back the utilities updates one by one when I'm back in the office tomorrow.   
After taking another look at my comment in https://github.com/GEOS-ESM/GEOSldas/issues/713, I realize I might have been to optimistic and may not have run the debug tests because those are model-only simulations and don't use HDF5.  My thinking was probably that we're ok as long as the standard optimization works for Intel and GNU with the "assim" test case that exercises HDF5.  But there may be some other change in ESMA_env that broke the debug simulation -- which is likely poor coding in GEOSldas revealed by debug.  Oh my.

biljanaorescanin:
All tests passed.

biljanaorescanin (approved):

PR #733, opened 2024-03-18
Allow users to setup a GEOSldas experiment from somebody else's build 

### LINKED ISSUE #699: run ./ldas_setup from another user's install
@weiyuan-jiang: `ldas_setup` must be run from the `./install/bin/` directory.  It appears that `ldas_setup` creates at least one temporary file (`LDAS_domain_def.nml`) in the `./install/bin/` directory:

https://github.com/GEOS-ESM/GEOSldas/blob/23f8320dac342b77ae6443a3f213fc5476aa240b/src/Applications/LDAS_App/ldas_setup#L361

This fails when the user who runs `ldas_setup` does not have write permission in the `./install/bin` directory, which normally happens when the user sets up an experiment from somebody else's build.  

Using somebody else's build for setting up an experiment works for GEOSadas and can be helpful in some situations.

Granting write permissions in `./install/bin/` to "group" or "other" is not an acceptable permanent solution.

There may be other temporary files written in `./install/bin/`.  

Can we make `ldas_setup` write such temporary files into a user-owned directory instead of `./install/bin/`?  If this is done after ldas_setup queries for the experiment directory, perhaps the temporary file could be created in the scratch directory of the experiment directory. 

cc: @saraqzhang @biljanaorescanin

mathomp4:
Closes #699 

This is my attempt at making `ldas_setup.py` usable by anyone. This (I think) should use a tempfile instead. Perhaps @biljanaorescanin can try it? Well I guess there are two tests:

- [x] Builder of the clone can make an experiment
- [x] Someone else can make an experiment using another's install dir

ETA: Note: my editor strips whitespace, so probably good to "hide whitespace" in the Files changed tab.


weiyuan-jiang:
Thanks @mathomp4 . This is a great start. But I believe it is not enough. The fortran program will produce a temporary file "f2g.txt".  I am wondering if fortran can accept it and write to it.  I will try.

mathomp4:
> Thanks @mathomp4 . This is a great start. But I believe it is not enough. The fortran program will produce a temporary file "f2g.txt". I am wondering if fortran can accept it and write to it. I will try.

Ohh. I didn't know about `f2g.txt`. Can the commands that need that file accept an input/output file name? Then I suppose you could make another NamedTemporaryFile, and pass that filename into the the commands that need it. Then it would all be temp files.

gmao-rreichle:
I was able to verify that I can setup and run an experiment using Weiyuan's build.
I also ran the nightly tests, which all passed except for the expected comparison "failure" for the new 6-h inst output in the LDAS_*GLOBAL/model tests.

gmao-rreichle (approved):

PR #739, opened 2024-03-21
Update some GitHub Actions versions

mathomp4:
Was doing some comparing and found that GEOSldas was still using `checkout@v3` in some GitHub Actions. The latest version is `v4` so we move to use that.

If the CI works, it's good code, if it doesn't it's bad code. GEOSldas itself doesn't care. 

Note: This is not a huge rush to get in.

### DIFF
--- a/.github/workflows/push-to-develop.yml
+++ b/.github/workflows/push-to-develop.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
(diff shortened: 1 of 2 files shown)


tclune (approved):

gmao-rreichle (approved):
Fine by me.  Probably best to merge immediately, so it'll be capture in the release I'm about to create.

mathomp4:
> Fine by me. Probably best to merge immediately, so it'll be capture in the release I'm about to create.

I pressed the merge button!

PR #796, opened 2025-02-26
Update env, cmake, GEOS_Util and MAPL releases in components.yaml & update README.md after decommissioning of SLES12 at NCCS

mathomp4:
This **non-0-diff** PR updates the `components.yaml` to approximately match that of GEOSgcm `main` as of 2025-Mar-19.  

```
env:       v4.29.1 --> v4.36.0
cmake:     v3.52.0 --> v3.57.0
GEOS_Util: v2.1.3  --> v2.1.6
MAPL:      v2.50.1 --> v2.54.2
```

The **non-0-diff** changes are within **"roundoff,"** and are caused by the newer compiler/baselibs version.  Intel tests with standard optimization are **0-diff when bit shaving is _not_ used**. 

Note that the ESMA_env, ESMA_cmake, and MAPL versions are slightly newer than those of GEOSgcm `main`, but per the respective release notes this should be 0-diff w.r.t. what is in GEOSgcm `main` (but **not** 0-diff w.r.t. what is on GEOSldas `develop` before this PR!).

The PR also updates README.md to reflect that SLES15 is now the only O/S on the NCCS Discover platform.

Earlier versions of this PR also included the helfsurface() optimization of https://github.com/GEOS-ESM/GMAO_Shared/pull/348, which requires a newer Intel compiler but should be zero-diff (which is why it will be done in a separate PR).

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,13 +45,13 @@ git clone -b v17.9.1 --single-branch git@github.com:GEOS-ESM/GEOSldas.git
 To build the model in a single step, do the following from a head node:
 ```
 cd ./GEOSldas
-./parallel_build.csh -mil
+./parallel_build.csh
 ```
 This checks out all the external repositories of the model (albeit only on the first run, [see subsection on mepo below](#mepo)!) and then builds and installs the model. 
 
 The resulting model build is found in `build-SLES15/`, and the installation is found in `install-SLES15/`, with setup scripts like `ldas_setup` in `install-SLES15/bin`.
 
-To obtain a build that is suitable for debugging, use `./parallel_build.csh -debug -mil`, which builds in `build-Debug-SLES15/` and installs in `install-Debug-SLES15/`.  There is also an option for aggressive  optimization.  For details, see the [GEOSldas Wiki](https://github.com/GEOS-ESM/GEOSldas/wiki).
+To obtain a build that is suitable for debugging, use `./parallel_build.csh -debug`, which builds in `build-Debug/` and installs in `install-Debug/`.  There is also an option for aggressive  optimization.  For details, see the [GEOSldas Wiki](https://github.com/GEOS-ESM/GEOSldas/wiki).
 
 Instructions for building the model in multiple steps are provided below.
 
@@ -60,9 +60,7 @@ Instructions for building the model in multiple steps are provided below.
 ## How to Set Up (Configure) and Run GEOSldas
 
 
-a) To run GEOSldas on Milan nodes (SLES15), start with `ssh discover-mil`.
-
-b) Set up the job as follows:
+Set up the job as follows:
 
 ```
 cd (build_path)/GEOSldas/install-SLES15/bin
--- a/README.md
+++ b/README.md
@@ -67,7 +67,7 @@ Instructions for building the model in multiple steps are provided below.
 Set up the job as follows:
 
 ```
-cd (build_path)/GEOSldas/install-SLES15/bin
+cd (build_path)/GEOSldas/install-Release/bin
 source g5_modules                                              [for bash or zsh: source g5_modules.[z]sh]
 ./ldas_setup setup [-v]  (exp_path)  ("exe"_input_filename)  ("bat"_input_filename)
 ```
--- a/README.md
+++ b/README.md
@@ -38,7 +38,11 @@ For science runs, you can also obtain a specific tag or branch _only_ (as oppose
 ```
 git clone -b v17.9.1 --single-branch git@github.com:GEOS-ESM/GEOSldas.git
 ```
-
+Helpful tip: You can speed up this step by applying the following, one-time `mepo` configuration change:
+```
+mepo config set clone.partial blobless
+```
+With this configuration change, the size of the initial clone will be reduced, and additional blobs are fetched later when needed.
 
 ### Step 3: Build the Model
 
@@ -49,7 +53,7 @@ cd ./GEOSldas
 ```
 This checks out all the external repositories of the model (albeit only on the first run, [see subsection on mepo below](#mepo)!) and then builds and installs the model. 
 
-The resulting model build is found in `build-SLES15/`, and the installation is found in `install-SLES15/`, with setup scripts like `ldas_setup` in `install-SLES15/bin`.
+The resulting model build is found in `build-Release/`, and the installation is found in `install-Release/`, with setup scripts like `ldas_setup` in `install-Release/bin/`.
 
 To obtain a build that is suitable for debugging, use `./parallel_build.csh -debug`, which builds in `build-Debug/` and installs in `install-Debug/`.  There is also an option for aggressive  optimization.  For details, see the [GEOSldas Wiki](https://github.com/GEOS-ESM/GEOSldas/wiki).
 
@@ -64,7 +68,7 @@ Set up the job as follows:
 
 ```
 cd (build_path)/GEOSldas/install-SLES15/bin
-source g5_modules                        [for bash or zsh: source g5_modules.[z]sh]
+source g5_modules                                              [for bash or zsh: source g5_modules.[z]sh]
 ./ldas_setup setup [-v]  (exp_path)  ("exe"_input_filename)  ("bat"_input_filename)
 ```
 
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ cd ./GEOSldas
 ```
 This checks out all the external repositories of the model (albeit only on the first run, [see subsection on mepo below](#mepo)!) and then builds and installs the model. 
 
-The resulting model build is found in `build-Release/`, and the installation is found in `install-Release/`, with setup scripts like `ldas_setup` in `install-Release/bin/`.
+The resulting model build is found in `build/`, and the installation is found in `install/`, with setup scripts like `ldas_setup` in `install/bin/`.
 
 To obtain a build that is suitable for debugging, use `./parallel_build.csh -debug`, which builds in `build-Debug/` and installs in `install-Debug/`.  There is also an option for aggressive  optimization.  For details, see the [GEOSldas Wiki](https://github.com/GEOS-ESM/GEOSldas/wiki).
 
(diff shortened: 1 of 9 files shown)


biljanaorescanin:
Testing summary: 
Almost all tests comparison will fail. All differences are roundoff. 
- Baselibs change from GCC 12 to GCC 14 the reason for all GNU fails. 
-  Intel fails are from new compilers. 

`Runtype         Clone Build Build Time Model Run/Compare Assim Run/Compare

conus           pass  pass  13 min     pass/FAIL          -- / --
global           --    --    --        pass/FAIL         pass/pass
globalcs         --    --    --        pass/FAIL         pass/pass
globalcnclm4     --    --    --        pass/FAIL          -- / --
debugconus       --   pass  11 min     pass/pass          -- / --
aggconus         --   pass  14 min     pass/FAIL          -- / --
aggglobal        --    --    --        pass/FAIL         pass/FAIL
aggglobalcs      --    --    --        pass/FAIL         pass/FAIL
aggglobalcnclm4  --    --    --        pass/FAIL          -- / --
gnuconus        pass  pass  31 min     pass/FAIL          -- / --
gnuglobal        --    --    --        pass/FAIL         pass/FAIL
gnuglobalcs      --    --    --        pass/FAIL         pass/FAIL
gnuglobalcnclm4  --    --    --        pass/FAIL          -- / --
gnudebugconus    --   pass  20 min     pass/pass          -- / --` 

Note: Helfand is not used as option during testing ( we use Louis as default) so for PR this change to use branch is trivially zero diff.

mathomp4:
Hmm. You are getting failures from the new Inte?

gmao-rreichle:
@biljanaorescanin, @mathomp4, 

I'm a bit confused by this PR and think we need to separate the update of the environment and the Helfand update.  Specifically:

1. It is not correct that only Louis is used.  The GLOBALCS/assim tests use Helfand:
```
LDAS_AGGGLOBALCS/assim/CURRENT/run/LDAS.rc:CHOOSEMOSFC:                        1
LDAS_GLOBALCS/assim/CURRENT/run/LDAS.rc:CHOOSEMOSFC:                        1
LDAS_GNUGLOBALCS/assim/CURRENT/run/LDAS.rc:CHOOSEMOSFC:                        1
```

2. I'm surprised that the comparison passes for the GLOBAL/assim test but fails for the GLOBAL/model test (and similarly for other tests).   

I think it would be best to remove the Helfand branch from the PR and examine the impact of the environment update in isolation.

biljanaorescanin:
1. Running without helfand branch is zero diff to running with helfand branch for intel tests confirming again helfand vectorization is zero diff.  
2. You will see below GNU tests fails and that is because for tests to pass we need baselibs GCC 14. 
     For my previous test summary @mathomp4 changed his sandbox to GCC 14. 
Screenshot 2025-03-11 at 12 35 53 PM

gmao-rreichle:
Thanks, @biljanaorescanin.  Here are my 2c:

> Running without helfand branch is zero diff to running with helfand branch for intel tests confirming again helfand vectorization is zero diff.

This is great, but I still think we want this to be in a separate PR for clarity.  When releases are made, the release doc is basically a collection of PR titles.  Having a separate PR for the zero-diff helfsurface() optimization change makes it much easier to understand what was done when a few months have passed and nobody can remember off the top of their head.  I edited the present PR accordingly.  Once the present PR has been merged, we can create, re-test, and merge a separate PR for the helfsurface() change.

> You will see below GNU tests fails and I think that is because for tests to pass we need baselibs GCC 14.  Maybe it can also be a discover glitch. Not 100% sure now since build failed.
> For my previous test summary @mathomp4 changed his sandbox to GCC 14.

What does it take to include the GCC-14 change into this PR?  It doesn't make sense to me to merge this PR when it doesn't work with the current GNU version.  Maybe I'm missing something. 

Also, I'm still surprised that the comparison passes for the GLOBAL/assim test but fails for the GLOBAL/model test (and similarly for other tests).  This could be a difference in 1d (tile) vs. 2d output and MAPL HISTORY regridding.  Before we can merge the PR, we need to understand better what exactly is not zero-diff here.

biljanaorescanin:
If I only focus to intel you will see only NC4 files fail and it is for roundoff: 
`conus/cmp_model.conus.log:Exception: Comparing outputs failed! BIN: True, `NC4: False`, RST: True
global/cmp_model.global.log:Exception: Comparing outputs failed! BIN: True, `NC4: False`, RST: True
globalcnclm4/cmp_model.globalcnclm4.log:Exception: Comparing outputs failed! BIN: True, `NC4: False`, RST: True
globalcs/cmp_model.globalcs.log: Comparing outputs failed! BIN: True, `NC4: False`, RST: True`

mathomp4:
> Thanks, @biljanaorescanin. Here are my 2c:
> > You will see below GNU tests fails and I think that is because for tests to pass we need baselibs GCC 14.  Maybe it can also be a discover glitch. Not 100% sure now since build failed.
> > For my previous test summary @mathomp4 changed his sandbox to GCC 14.
> 
> What does it take to include the GCC-14 change into this PR? It doesn't make sense to me to merge this PR when it doesn't work with the current GNU version. Maybe I'm missing something.

This is actually an issue with the *scripting*. In the regression scripts, for GNU runs I have to replace the `g5_modules` with one for GNU. At the moment, the LDAS scripting still uses a GCC 12 `g5_modules` (since moving to GCC 14 would be non-zero-diff).

I can change that in the scripting, and then the GNU tests would go NZD the next time things run.

README.md
```diff
@@ -39,22 +38,24 @@ For science runs, you can also obtain a specific tag or branch _only_ (as oppose
...
 To build the model in a single step, do the following from a head node:
 ```
 cd ./GEOSldas
-parallel_build.csh
+./parallel_build.csh
 ```
 This checks out all the external repositories of the model (albeit only on the first run, [see subsection on mepo below](#mepo)!) and then builds and installs the model. 
 
-At **NCCS**, the default is to build GEOSldas on SLES12 (Skylake or Cascade Lake nodes); to build GEOSldas on SLES15 (Milan nodes), use `parallel_build.csh -mil`.
+The resulting model build is found in `build-Release/`, and the installation is found in `install-Release/`, with setup scripts like `ldas_setup` in `install-Release/bin/`.
```
gmao-rreichle, line 56: @weiyuan-jiang: I edited this paragraph and the next (https://github.com/GEOS-ESM/GEOSldas/pull/796/commits/1dba0d7e786250f5279f26c4b539b1cf2905fcbe) based on the build of the LDAS nightly tests.  Please verify if this matches what you get when you build the model.

components.yaml
```diff
@@ -5,13 +5,13 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v4.29.1
+  tag: v4.36.0
```
gmao-rreichle, line 8: @biljanaorescanin, @mathomp4  : I ticked up the versions of  env, cmake, and MAPL (https://github.com/GEOS-ESM/GEOSldas/pull/796/commits/c1007c887e99f3dccba835e56fba01c80dad9b58).  Based on the documentation of the respective releases, this should be zero-diff w.r.t. what was on the PR before my latest edits (but definitely non-0-diff w.r.t. current develop).  @mathomp4, please let me know if you have any objections or suggestions.  @biljanaorescanin, when you get a chance, please re-test the PR.  If all is as expected, the new test is 0-diff w.r.t. the most recent test (if you still have a copy).

gmao-rreichle (commented):
See inline comments below.

biljanaorescanin:
Tests are zero diff to previous iteration of testing. 
 We don't get GNU tests since I didn't use right baselibs that needs to be changed on Matt's side, but we do have results from before if anyone want's to take a look it was roundoff difference.

gmao-rreichle:
@mathomp4, @biljanaorescanin, @weiyuan-jiang: 

I am still trying to understand the very unusual non-0-diff character of this PR.  Specifically, the LDAS_GLOBAL/model test fails the comparison for the nc4 files (in just a small subset of variables, and within what seems to be roundoff).  The curious thing is that the LDAS_GLOBAL/**assim** test **passes**!  

If there was any change in the science code (or a roundoff change in the science calcs triggered by the newer env/baselibs), then the assim test should also fail the nc4 comparison.   The fact that the assim test passes suggests that it's something in MAPL and/or the LDAS tile_bin2nc4 utility.  That is, the variables would need to be 0-diff when they in memory during the simulation, but then something changes when the data are written out. 

I noticed that the Intel tests with standard optimization that do pass have no bit shaving in HISTORY.rc, whereas the tests that fail the nc4 comparison have bit shaving enabled.  I went through the documentation of the MAPL releases between 2.50.1 and 2.54.2 and didn't notice anything that might have impacted the bit shaving, and the documentation suggests that for the most part the MAPL releases in question should all be 0-diff among themselves (for the GCM, which is usually a bigger hurdle than LDAS when it comes to 0-diff).  So I can't see how exactly the bit shaving might cause the non-0-diffs seen here, but I also can't quite rule it out.  

Thoughts?

weiyuan-jiang:
The failed comparison in model run is on files  that are not in the assim run. So we probably don't need to worry about this.

gmao-rreichle:
> The failed comparison in model run is on files that are not in the assim run. So we probably don't need to worry about this.

@weiyuan-jiang, I'm not sure I understand the reasoning.  Of course the "model" test case has different HISTORY output.  What I'm after is understanding which exact changes in the PR caused the non-0-diff result for the model test case.  Normally, anything that causes non-0-diff in the model test case would also cause non-0-diff in the assim test case.  The fact that the model case is non-0-diff but the assim case is 0-diff is **very** unusual, and I'd really like to be able to explain this so we can make more informed decisions about how to interpret the non-0-diff changes in science applications going forward

gmao-rreichle:
For lack of a better idea I just tested a variant of the PR's branch that reverts MAPL back to 2.50.1.  I only ran the Intel tests w/ standard optimization.  The result is 0-diff w.r.t. using MAPL 2.54.2, so MAPL is **not** the cause of the non-0-diff result vs. develop.  Still a mystery to me why we get non-0-diff for output from the "model" test but 0-diff for the "assim" test.

biljanaorescanin:
In our regression testing we got test fail:  `global           --    --    --        pass/FAIL         pass/pass `

If I run just GLOBAL/model test and comment out in `HISTORY.rc`  `*.nbits:       12,` so we are not using bit shaving,  then our develop vs branch comparison is **zero diff** for both collection comparisons.  
In our GLOBAL/assim run we don't use bit shaving in history.rc and that is why run was a pass for comparison.

gmao-rreichle (approved):

PR #830, opened 2025-08-14
WIP: Testing 8.17 AEC Baselibs

mathomp4:
I am testing a new Baselibs that moves away from using szip which has long been considered ancient and recently the HDF Group just *removed* the page about it on their site. Instead, it moves to libaec which is the "successor" of szip and can be used the same, just with some link-time changes.

The only reason I'm concerned with the LDAS is that it might be one of the places we still use szip. I mean, GEOSldas assimilates HDF4 files and maybe those are szip compressed? I dunno.

So, I'm hoping someone can run the Intel tests, especially assim, with this PR. Just to see.

I pinged @weiyuan-jiang on Teams to ask if he could run it. I don't want to make a mistake and screw up my nightly tests somehow.

### DIFF
--- a/components.yaml
+++ b/components.yaml
@@ -57,7 +57,7 @@ MAPL:
 GEOSldas_GridComp:
   local: ./src/Components/@GEOSldas_GridComp
   remote: ../GEOSldas_GridComp.git
-  branch: develop
+  branch: testing-8.17-AEC-baselibs
   develop: develop
 
 GEOSgcm_GridComp:
--- a/components.yaml
+++ b/components.yaml
@@ -11,7 +11,7 @@ env:
 cmake:
   local: ./@cmake
   remote: ../ESMA_cmake.git
-  tag: v3.62.1
+  branch: feature/v3-add-aec-support
   develop: develop
 
 ecbuild:
--- a/components.yaml
+++ b/components.yaml
@@ -5,7 +5,7 @@ GEOSldas:
 env:
   local: ./@env
   remote: ../ESMA_env.git
-  tag: v4.38.0
+  branch: testing-8.17-AEC-baselibs
   develop: main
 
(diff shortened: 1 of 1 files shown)


mathomp4:
Note: As far as I can tell, the file it failed on is not compressed with szip but with boring old zlib.

@gmao-rreichle Do you know if any files the LDAS assimilates are compressed with szip?

gmao-rreichle:
> Note: As far as I can tell, the file it failed on is not compressed with szip but with boring old zlib.
> 
> @gmao-rreichle Do you know if any files the LDAS assimilates are compressed with szip?

@mathomp4 : Apologies. I hadn't even heard of all these flavors of compression until you created the PR.  I agree that it's odd that the read error happens on the 2nd file, and I don't have a clue why this would be.

gmao-rreichle:
> GNU complier passed all the tests . I have forwarded you the email

Well, the Baselibs don't control the GNU compiler, if I understand correctly.  The GNU version is built into the regression script somehow, I think.  So no surprise there. 

Otherwise, the model tests all run, which is good, and the standard Intel builds also pass the test.

But none of the Intel assim tests run, confirming the issue with reading the SMAP h5 files that Matt mentioned above.  

I have no clue why there's an issue with reading the h5 files

mathomp4:
> @mathomp4 : Now that we merged [GEOS-ESM/GEOSldas_GridComp#139](https://github.com/GEOS-ESM/GEOSldas_GridComp/pull/139), should I re-test this PR? Are the new baselibs branches in env and cmake still ok? I assume we need to revert the GEOSldas_GridComp branch to "develop".

@gmao-rreichle Let me close this. I'm making tags of various repos, etc. So then I can be a bit more exact with the test.

PR #836, opened 2025-09-19
Update components to match GEOSgcm v11 as of 2025-Sep-19

mathomp4:
This PR updates the components in GEOSldas to match that of GEOSgcm `main` as of 2025-Sep-19. This brings in a few things:

* ESMA_env v5.13.0
  * Update to Baselibs 8.18.0
    * ESMF 8.9.0
    * curl 8.15.0
    * NCO 5.3.4
    * CDO 2.5.3
    * nccmp 1.10.0.0
    * Removed szip, added libaec
      * Note: To use libaec correctly, users should use ESMA_cmake v3.63.0/v4.20.0 or higher
* ESMA_cmake v3.64.0
  * Support for libaec in Baselibs
* GMAO_Shared v2.1.4
  * Changes to some scripts and `esma_mpirun`
* GEOS_Util v2.1.10
  * Various updates (see https://github.com/GEOS-ESM/GEOS_Util/blob/main/CHANGELOG.md)
* MAPL 2.59.1
  * Support for ESMF v9
  * Various fixes and features

I've also updated some CI to test if all this works.

The *most* important part of this is MAPL 2.59.1. That is important because it provides support for ESMF v9. Now, MAPL2 (and thus GEOSgcm, GEOSldas, etc.) does *not* require ESMF v9 but MAPL3 does. But, having support to use ESMF v9 allows us to keep around a single set of Baselibs/ESMA_env.

Second, #835 by @biljanaorescanin trumps this in terms of GEOS_Util, but I decided to just echo GEOSgcm `main`

Third, as I know @gmao-rreichle will probably spot it, ESMA_env went from v4 to v5. Now, that difference is because ESMA_env v5 has Baselibs v8 which includes FMS (which doesn't matter to GEOSldas).

I'm going to keep this draft until it can be tested by @biljanaorescanin or others. But I wanted to get this in to see what CI does.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,7 +1,7 @@
 version: 2.1
 
 # Anchors in case we need to override the defaults from the orb
-#baselibs_version: &baselibs_version v7.33.0
+#baselibs_version: &baselibs_version v8.18.0
 #bcs_version: &bcs_version v11.6.0
 
 orbs:
--- a/.github/workflows/release-tarball.yml
+++ b/.github/workflows/release-tarball.yml
@@ -10,12 +10,12 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v5
         with:
           path: ${{ github.event.repository.name }}-${{ github.event.release.tag_name }}
 
       - name: Setup Python
-        uses: actions/setup-python@v5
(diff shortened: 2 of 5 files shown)


mathomp4:
> All tests passed.

Hooray! Thanks for testing. I'll undraft.

